PR #10, opened 2020-03-02
Unaligned support

MushMal:
* Unaligned support in PIC
* Content store -> AIV Heap -> by default, will be tightly packed in the library and aligned in the unit tests. 
* Enabled the tests which were disabled previously
* Enabled UBSAN with a single exception for MAX INT64 in string conversion test (BY DESIGN)


src/heap/tst/HeapApiFunctionalityTest.cpp
```diff
@@ -136,7 +143,7 @@ VOID minBlockFitAllocResize(PHeap pHeap)
     UINT64 retSize, setSize;
 
     // Set to default
-    MEMSET(handles, INVALID_ALLOCATION_HANDLE_VALUE, NUM_ITERATIONS);
+    zeroHandleArray(handles, NUM_ITERATIONS);
```
disa6302: Why are we replacing `MEMSET()` with `zeroHandleArray()`?

`MEMSET(handles, INVALID_ALLOCATION_HANDLE_VALUE, NUM_ITERATIONS * SIZEOF(ALLOCATION_HANDLE)` should do it, right? DO we need `zeroHandleArray()`?

src/client/tst/StreamPutGetTest.cpp
```diff
@@ -1518,7 +1518,7 @@ TEST_F(StreamPutGetTest, DISABLED_putFrame_PutGetTagsBeforeStoreData)
     ASSERT_TRUE(retStatus == STATUS_SUCCESS || retStatus == STATUS_NO_MORE_DATA_AVAILABLE);
 
     // Manually pre-validated data file size
-    EXPECT_EQ(6700, filledSize);
+    EXPECT_EQ(6703, filledSize);
```
disa6302: Can we have a #define for 6703? It looks like a magic number and having a #define will be useful for understanding in the future.

src/heap/src/Common.h
```diff
@@ -9,30 +9,49 @@
...
-#pragma once
-#pragma pack(push, include, 1) // for byte alignment
-
 /**
  * 'Magic' - aka a sentinel which will be added as a guard band for each allocation in debug mode
  */
 #define ALLOCATION_HEADER_MAGIC "__HEADER_MAGIC_GUARD_BAND__"
 #define ALLOCATION_FOOTER_MAGIC "__FOOTER_MAGIC_GUARD_BAND__"
 
+// Ensure it's more than the actual header and footer size
```
zhiyua-git: It would be clearer to say more than size of ALLOCATION_HEADER_MAGIC and ALLOCATION_FOOTER_MAGIC

src/heap/src/AivHeap.c
```diff
@@ -6,11 +6,11 @@
 #include "Include_i.h"
 
 #ifdef HEAP_DEBUG
-    AIV_ALLOCATION_HEADER gAivHeader = {{0, AIV_ALLOCATION_TYPE, 0, ALLOCATION_HEADER_MAGIC}, 0, ALLOCATION_FLAGS_FREE, NULL, NULL};
-    AIV_ALLOCATION_FOOTER gAivFooter = {{0, ALLOCATION_FOOTER_MAGIC}};
+    struct AIV_ALLOCATION_HEADER gAivHeader = {{0, AIV_ALLOCATION_TYPE, ALLOCATION_FLAGS_FREE, ALLOCATION_HEADER_MAGIC}, 0, NULL, NULL};
```
disa6302: Just a thought, but, can we define the values for every field using the . operator? It will just be safer to do so, in case somebody changes the order of structure items in the future?

src/heap/tst/HeapApiFunctionalityTest.cpp
```diff
@@ -420,11 +428,23 @@ TEST_F(HeapApiFunctionalityTest, AivHeapMinBlockFitResize)
...
+    EXPECT_EQ(MIN_HEAP_SIZE + 1, pHeap->heapLimit);
+#endif
+}
+
 TEST_F(HeapApiFunctionalityTest, AivHeapResizeEdgeCases)
 {
     PHeap pHeap;
     ALLOCATION_HANDLE handle = INVALID_ALLOCATION_HANDLE_VALUE;
-    UINT32 size = MIN_HEAP_SIZE / NUM_ITERATIONS;
+    UINT32 size = ROUND_UP(MIN_HEAP_SIZE / NUM_ITERATIONS, 8);
```
zhiyua-git: Would it be better to use HEAP_PACKED_SIZE?

src/utils/src/Tags.c
```diff
@@ -75,9 +75,13 @@ STATUS packageTags(UINT32 tagCount, PTag pSrcTags, UINT32 tagsSize, PTag pDstTag
         curSize += nameSize + valueSize;
     }
 
+    // Round-up the allocation size to the 64 bit
+    curSize = ROUND_UP(curSize, 8);
```
zhiyua-git: Should be HEAP_PACKED_SIZE()?

src/heap/src/SystemHeap.c
```diff
@@ -321,6 +322,11 @@ DEFINE_FOOTER_SIZE(sysGetAllocationFooterSize)
     return SYS_ALLOCATION_FOOTER_SIZE;
 }
 
+DEFINE_ALIGNED_SIZE(sysGetAllocationAlignedSize)
```
disa6302: Looks like the macro name is same for System Heap as well as Hybrid heap. Just for sake of clarity, can we rename these to indicate the type of heaps? `DEFINE_ALIGNED_SIZE_SYSTEM_HEAP` probably?

src/heap/tst/HeapApiFunctionalityTest.cpp
MushMal: The main issue is that MEMSET operates on byte granularity. INVALID_ALLOCATION_HANDLE_VALUE is evaluated to UINT64. We can't assume that the value is 0 - it can be any arbitrary number. Imagine if invalid handle value was 0x01020304050507. Using MEMSET would set the entire array to 0x07

src/client/tst/StreamPutGetTest.cpp
MushMal: This would have been absolutely a no go for the codebase but the tests assume these numbers have been sanitized by hand (which we do by running mkvgen -v). As the test code is peppered with these, I won't go and change these now - we can scrub them at a later stage if needed.

src/heap/src/AivHeap.c
MushMal: These are static definitions so you can't dynamically execute code. Moreover, this is C and not C++. I would imagine anyone changing a piece of code at this layer would be super familiar with what's going on.

CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ option(ADDRESS_SANITIZER "Build with AddressSanitizer." OFF)
 option(MEMORY_SANITIZER "Build with MemorySanitizer." OFF)
 option(THREAD_SANITIZER "Build with ThreadSanitizer." OFF)
 option(UNDEFINED_BEHAVIOR_SANITIZER "Build with UndefinedBehaviorSanitizer." OFF)
+option(HEAP_PERF_TESTS "Build with heap perf tests." OFF)
+option(BUILD_DEBUG_HEAP "Build debug heap with guard bands and validation." OFF)
+option(ALIGNED_MEMORY_MODEL "Aligned memory model ONLY." OFF)
```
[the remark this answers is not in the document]
MushMal: Don't think so. 

HEAP_PERF_TESTS run for a while and might produce weird results in Travis environment.
BUILD_DEBUG_HEAP will run two orders of magnitude slower and will fail tests due to bloat size of the allocations - this is expected
ALIGNED_MEMORY_MODEL - most devices will not need aligned access on the heap and will save on memory and performance. The tests, on the other hand, are compiled with this flag on.

CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ option(ADDRESS_SANITIZER "Build with AddressSanitizer." OFF)
 option(MEMORY_SANITIZER "Build with MemorySanitizer." OFF)
 option(THREAD_SANITIZER "Build with ThreadSanitizer." OFF)
 option(UNDEFINED_BEHAVIOR_SANITIZER "Build with UndefinedBehaviorSanitizer." OFF)
+option(HEAP_PERF_TESTS "Build with heap perf tests." OFF)
```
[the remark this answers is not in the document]
MushMal: For our sanity validation/checks. It's going to produce untrusted results on travis.

CMakeLists.txt
```diff
@@ -143,6 +156,15 @@ if(BUILD_TEST)
...
+  add_executable(kvspic_test ${PIC_TEST_SOURCE_FILES}
+          ${PIC_CLIENT_SOURCE_FILES}
+          ${PIC_DURATION_SOURCE_FILES}
+          ${PIC_HEAP_SOURCE_FILES}
+          ${PIC_MKVGEN_SOURCE_FILES}
+          ${PIC_STATE_SOURCE_FILES}
+          ${PIC_TRACE_SOURCE_FILES}
+          ${PIC_UTILS_SOURCE_FILES}
+          ${PIC_VIEW_SOURCE_FILES})
+  target_compile_definitions(kvspic_test PRIVATE ALIGNED_MEMORY_MODEL=TRUE)
```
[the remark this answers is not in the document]
MushMal: We need the tests to run with aligned model to ensure the ubsan runs clean. We can have another run with tight packed heap tests and no UBSAN - let me create an issue for it: https://github.com/awslabs/amazon-kinesis-video-streams-pic/issues/11

src/client/tst/StreamPutGetTest.cpp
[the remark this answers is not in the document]
MushMal: I can't quite understand how this passed before. Seems proper size and validated with mkvgen -v

src/heap/src/Common.h
MushMal: These are called guard bands in allocators

CMakeLists.txt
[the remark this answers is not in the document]
zhiyua-git: This comment is for ALIGNED_MEMORY_MODEL. Just wonder if we should just provide customer with default aligned access. With this being OFF, they will still need to contact us for help when they hit bus error on the device when they put the code on different platforms, but if it is ON, most of these failures can be avoid. It might be OK to note down the solution in trouble shooting guide, but I wonder if default to be ON would make life easier for both us and them.

CMakeLists.txt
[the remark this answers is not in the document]
zhiyua-git: If we don't regularly run this on travis, when is it run?

src/heap/tst/HeapApiFunctionalityTest.cpp
MushMal: I think it's a good idea - though, we still would fail if the heap and the tests are compiled differently. Will change

src/utils/src/Tags.c
MushMal: No - these are always aligned unlike the heap

src/heap/src/SystemHeap.c
MushMal: These are helper macros that define the functions for various implementations. Currently, we have 3 implementations and all would use these macros to define the same functions. Then, in the Heap.c we implement the polymorphism - this is the pattern we use in multiple other places to achieve polymorphism without expensive v-tables.

CMakeLists.txt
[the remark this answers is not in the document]
MushMal: We need to take this offline to discuss the canaries, sanity runs in Jenkins

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=h1) Report
> Merging [#10](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/commit/433f3b35ee8cc94522a6ea6bc7bb3ca14e576fd9&el=desc) will **decrease** coverage by `4.23%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master      #10      +/-   ##
==========================================
- Coverage   85.48%   81.24%   -4.24%     
==========================================
  Files          45       47       +2     
  Lines        7985     8206     +221     
==========================================
- Hits         6826     6667     -159     
- Misses       1159     1539     +380     
```


| [Impacted Files](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/utils/src/TimerQueue.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL3V0aWxzL3NyYy9UaW1lclF1ZXVlLmM=) | `0.00% <0.00%> (-82.81%)` | :arrow_down: |
| [src/utils/src/Endianness.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL3V0aWxzL3NyYy9FbmRpYW5uZXNzLmM=) | `68.14% <0.00%> (+16.97%)` | :arrow_up: |
| [src/utils/src/Thread.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL3V0aWxzL3NyYy9UaHJlYWQuYw==) | `52.54% <0.00%> (-10.17%)` | :arrow_down: |
| [src/utils/src/Hex.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL3V0aWxzL3NyYy9IZXguYw==) | `0.00% <0.00%> (ø)` | |
| [src/utils/src/Base64.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL3V0aWxzL3NyYy9CYXNlNjQuYw==) | `0.00% <0.00%> (ø)` | |
| [src/client/src/Client.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL2NsaWVudC9zcmMvQ2xpZW50LmM=) | `83.94% <0.00%> (+0.13%)` | :arrow_up: |
| [src/client/src/InputValidator.c](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10/diff?src=pr&el=tree#diff-c3JjL2NsaWVudC9zcmMvSW5wdXRWYWxpZGF0b3IuYw==) | `84.24% <0.00%> (+3.19%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=footer). Last update [433f3b3...c6f6e48](https://codecov.io/gh/awslabs/amazon-kinesis-video-streams-pic/pull/10?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

CMakeLists.txt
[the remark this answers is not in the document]
chehefen: We can just disable the test in travis like you did to timeQueueTests. There is not much point to build a cmake flag for this.

src/heap/src/Common.h
```diff
@@ -9,30 +9,49 @@
...
+
+    // The type of the heap
     UINT32 type;
-    UINT32 handle;
+
+    // Union used for flags or VRAM handler
+    union {
+        UINT32 vramHandle;
+        UINT32 flags;
+    };
```
chehefen: why use union for two variables of same type? Also it's not clear how vramHandle cant co-exist with flags.

MushMal: The main idea here is the field name. For VRAM-based heap, it will use vramHandle. For AIV heap, it will be flags. Later, for file, it will be segment or something.

CMakeLists.txt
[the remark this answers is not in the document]
MushMal: I guess we could do that too. Let me modify the code to get it into a separate file with it's name so we can disable all perf.

src/client/src/Client.c
```diff
@@ -93,11 +95,8 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
                          pKinesisVideoClient->clientCallbacks.customData);
     }
 
-    // Fix-up the defaults if needed
-    // IMPORTANT!!! The calloc allocator will zero the memory which will also act as a
-    // sentinel value in case of an earlier version of the structure
-    // is used and the remaining fields are not copied
-    fixupClientInfo(&pKinesisVideoClient->deviceInfo.clientInfo);
+    // Set logger log level
+    SET_LOGGER_LOG_LEVEL(pKinesisVideoClient->deviceInfo.clientInfo.loggerLogLevel);
```
chehefen: are we making the assumption that clientInfo exists? what happens if deviceInfo is older version.

src/heap/src/Common.c
```diff
@@ -231,7 +231,9 @@ DEFINE_HEAP_ALLOC(commonHeapAlloc)
     CHK_ERR(pHeap->heapLimit != 0, STATUS_HEAP_NOT_INITIALIZED, "Heap has not been initialized.");
 
     // Check if we can allocate the chunk taking into account the allocation header and footer
-    overallSize = pBaseHeap->getAllocationHeaderSizeFn() + size + pBaseHeap->getAllocationFooterSizeFn();
+    overallSize = pBaseHeap->getAllocationHeaderSizeFn(size) +
```
chehefen: are getAllocationHeaderSizeFn and getAllocationFooterSizeFn taking param?

chehefen:
one last thing, did we actually test this on a platform that dont support unaligned access?

CMakeLists.txt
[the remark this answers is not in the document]
MushMal: True but vast majority of the customers will not have degraded perf and more fragmentation

src/heap/src/Common.c
MushMal: Yes indeed, those don't take size. Will change

MushMal:
We have a clean UBSAN run. I don't anticipate issues with unaligned platforms but we should test on RPi with aligned only mode.

PR #16, opened 2020-03-07
add SerialDispatchQueue utility

chehefen:
*Issue #, if available:*

*Description of changes:*


src/utils/src/SerialDispatchQueue.c
```diff
@@ -0,0 +1,267 @@
...
+    CHK(NULL != (pSerialDispatchQueue = (PSerialDispatchQueue) MEMCALLOC(1, SIZEOF(SerialDispatchQueue))),
+        STATUS_NOT_ENOUGH_MEMORY);
+    pSerialDispatchQueue->lock = MUTEX_CREATE(TRUE);
+    CHK(IS_VALID_MUTEX_VALUE(pSerialDispatchQueue->lock), STATUS_INTERNAL_ERROR);
+    pSerialDispatchQueue->executorCvar = CVAR_CREATE();
+    CHK(IS_VALID_CVAR_VALUE(pSerialDispatchQueue->executorCvar), STATUS_INTERNAL_ERROR);
+    pSerialDispatchQueue->shutdown = FALSE;
+    CHK_STATUS(stackQueueCreate(&pSerialDispatchQueue->taskqueuePriDefault));
+    CHK_STATUS(stackQueueCreate(&pSerialDispatchQueue->taskqueuePriHigh));
+    THREAD_CREATE(&pSerialDispatchQueue->executorTid, serialDispatchQueueExecutor, (PVOID) pSerialDispatchQueue);
```
MushMal: THREAD_DETACH it too

src/utils/src/SerialDispatchQueue.c
```diff
@@ -0,0 +1,267 @@
...
+CleanUp:
+
+    LEAVES();
+    return retStatus;
+}
+
+STATUS serialDispatchQueueDispatchTask(SERIAL_DISPATCH_QUEUE_HANDLE handle,
+                                       DispatchQueueTaskFunc dispatchQueueTaskFn,
+                                       UINT64 customData,
+                                       DISPATCH_QUEUE_TASK_PRIORITY priority)
```
MushMal: To be consistent with other public APIs, make sure customData and dispatchQueuTaskFn in the right order. Also, move the priority as a second param - right after the handle.

src/utils/src/SerialDispatchQueue.c
```diff
@@ -0,0 +1,267 @@
...
+    PSerialDispatchQueue pSerialDispatchQueue = NULL;
+    BOOL locked = FALSE;
+    PDispatchQueueTask pDispatchQueueTask = NULL;
+
+    CHK(dispatchQueueTaskFn != NULL, STATUS_NULL_ARG);
+    CHK(IS_VALID_SERIAL_DISPATCH_QUEUE_HANDLE_VALUE(handle), STATUS_INVALID_ARG);
+
+    pSerialDispatchQueue = FROM_SERIAL_DISPATCH_QUEUE_HANDLE(handle);
+
+    CHK(!ATOMIC_LOAD_BOOL(&pSerialDispatchQueue->shutdown), retStatus);
```
MushMal: I think this should return an error

src/utils/src/SerialDispatchQueue.c
```diff
@@ -0,0 +1,267 @@
...
+
+    CHK_LOG_ERR(retStatus);
+
+    if (locked) {
+        MUTEX_UNLOCK(pSerialDispatchQueue->lock);
+    }
+
+    LEAVES();
+    return (PVOID) (ULONG_PTR) retStatus;
+}
```
MushMal: Append a \n at the end of the file - some sanitizers complain for some reason.

src/utils/tst/SerialDispatchQueue.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    appendAlphabetPriDefault.join();
+    appendAlphabetPriHigh.join();
+
+    THREAD_SLEEP(50 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+
+    EXPECT_TRUE(!STRCMP(alphabetPriDefault, "abcdefghijklmnopqrstuvwxyz"));
+    EXPECT_TRUE(!STRCMP(alphabetPriHigh, "abcdefghijklmnopqrstuvwxyz"));
+    EXPECT_EQ(STATUS_SUCCESS, serialDispatchQueueFree(&handle));
+}
```
MushMal: \n at the end of the file

MushMal:
One more comment about the executor thread. Perhaps, we should seriously consider having a ThreadPool primitive which then would power the task queue. This can be a step closer to a Vat

src/utils/src/SerialDispatchQueue.c
chehefen: I dont need to detach it as the dispatch queue own the executor thread and will join it when dispatch queu is freed.

chehefen:
> One more comment about the executor thread. Perhaps, we should seriously consider having a ThreadPool primitive which then would power the task queue. This can be a step closer to a Vat

The would be a concurrent dispatch queue. I begin with serial dispatch queue because some tasks that we want to execute need to be executed sequentially.

chehefen:
> This is a great start indeed. What I really love is the clearness and correctness of the code you've produced!
> 
> Here are some of the higher-level comments I would like to discuss...
> 
> Short term:
> Define a priority which is an unsigned integer in the range of 0 - MAX_PRI. 0 being the highest or lowest. Have a single task queue. Insertion should happen from the end up until the priority matches. This would ensure insertion to the last position of the known priority.
> Optimization: Small tasks would execute fast alloc/free of the task - this is a huge task for small tasks. As your tasks are fixed size, we should consider having an array of structures and a rolling window on top of it instead of StackQueue which is dynamic alloc.
> 
> Long term:
> As you implement the array based allocation (like the ContentView), you can think of implementing lockless execution. This is not too hard and will benefit greatly for high-frequency/high-volume task queue. We can discuss different approaches.

Regarding the priority, I don't have a long range of priorities will benefit us. When a task is created, it will be very hard to predict how the priority level will affect its time of execution. Apple's implementation only has 3 priority levels: low, default and high. 
Regarding optimization, I dont think we should do too much optimization right from the beginning. As soon as the api stabilize then we can optimize the internal implementation.

src/utils/tst/SerialDispatchQueue.cpp
chehefen: fixed

src/utils/src/SerialDispatchQueue.c
chehefen: fixed

src/utils/src/SerialDispatchQueue.c
chehefen: fixed

src/utils/src/SerialDispatchQueue.c
chehefen: I dont need to detach it. SerialThreadDispatchQueue will join it when the structure get freed.

PR #20, opened 2020-03-16
allow content view to drop entire fragment when overflown. Drop fragm…

chehefen:
…ent by default.

*Issue #, if available:*

*Description of changes:*


src/view/src/ContentView.c
```diff
@@ -462,19 +463,44 @@ STATUS contentViewAddItem(PContentView pContentView, UINT64 timestamp, UINT64 ac
...
+
+                        // Callback if it's specified
+                        if (pRollingView->removeCallbackFunc != NULL) {
+                            // NOTE: The call is prompt - shouldn't block
+                            pRollingView->removeCallbackFunc(pContentView, pRollingView->customData, pTail,
+                                                             !currentAvailability);
+                        }
+
+                        pTail = GET_VIEW_ITEM_FROM_INDEX(pRollingView, pRollingView->tail);
+                    } while (!CHECK_ITEM_FRAGMENT_START(pTail->flags) && pRollingView->tail != pRollingView->head);
```
MushMal: Could you double check whether you need to check CHECK_ITEM_STREAM_START or we always have both flags on the session start.

src/view/src/Include_i.h
```diff
@@ -24,6 +24,11 @@ extern "C" {
  */
 #define GET_VIEW_ITEM_FROM_INDEX(view, index) (&(view)->itemBuffer[((index) == 0) ? 0 : ((index) % (view)->itemBufferCount)])
 
+typedef enum {
```
MushMal: Good comments please

src/view/src/ContentView.c
```diff
@@ -40,6 +40,7 @@ STATUS createContentView(UINT32 maxItemCount, UINT64 bufferDuration, ContentView
     pContentView->removeCallbackFunc = removeCallbackFunc;
     pContentView->itemBufferCount = maxItemCount;
     pContentView->bufferDuration = bufferDuration;
+    pContentView->bufferOverflowStrategy = DROP_UNTIL_FRAGMENT_START;
```
MushMal: I am kind of split on this but we should perhaps take this flag in and make decisions about this in the client.

MushMal: Also, I don't think we need to provide backward compatibility for this as we have no clients using content view directly.

src/view/src/Include_i.h
```diff
@@ -60,6 +65,8 @@ typedef struct {
     // The actual buffer which follows immediately after the structure
     PViewItem itemBuffer;
 
+    CONTENT_VIEW_OVERFLOW_STRATEGY bufferOverflowStrategy;
```
MushMal: Super careful here! The entire structure and the array is crafted to be 64bit aligned. Move the bufferOverflowStrategy before itemBufferCount. Add comments. Make the variable UINT64 to not break the alignment. Cast-assign to UINT64. Another strategy is to use CONTENT_VIEW_OVERFLOW_STRATEGY type and pad with UINT32 reserved;

src/view/tst/ViewApiFunctionalityTest.cpp
```diff
@@ -295,6 +295,88 @@ TEST_F(ViewApiFunctionalityTest, AddGetSetCurrentRemoveAll)
...
+                                                 timestamp,
+                                                 timestamp,
+                                                 TEST_VIEW_ITEM_DURATION,
+                                                 INVALID_ALLOCATION_HANDLE_VALUE,
+                                                 0,
+                                                 VIEW_ITEM_ALLOCAITON_SIZE,
+                                                 i % KEY_FRAME_INTERVAL == 0 ? ITEM_FLAG_FRAGMENT_START : ITEM_FLAG_NONE));
+    // entire fragment should get dropped
+    EXPECT_EQ(gCallCount, KEY_FRAME_INTERVAL);
+}
```
MushMal: Could you add the following test cases 1) to drop entire content view on overflow with the new strategy and 2) drop until new session 3) new strategy with EoFR.

src/view/src/ContentView.c
chehefen: yes we always have fragment start.

sharadrj:
@chehefen : Is there any impact on AV Sync with this fragment drop logic? I am assuming not but just want to be sure.

src/view/tst/ViewApiFunctionalityTest.cpp
chehefen: test added.

src/client/src/Stream.c
```diff
@@ -229,6 +229,7 @@ STATUS createStream(PKinesisVideoClient pKinesisVideoClient, PStreamInfo pStream
                                  pKinesisVideoStream->streamInfo.streamCaps.bufferDuration,
                                  viewItemRemoved,
                                  TO_CUSTOM_DATA(pKinesisVideoStream),
+                                 DROP_UNTIL_FRAGMENT_START,
```
MushMal: I will take care of exposing this further in V2 of StreamCaps as I am doing the next step of the work on storage pressure

chehefen:
> @chehefen : Is there any impact on AV Sync with this fragment drop logic? I am assuming not but just want to be sure.

no, there should be no impact.

Issue #22, opened 2020-03-17
Enable tail eviction on the storage pressure


MushMal:
Seems that some of the customers would like to have a behavior controlled by a flag that will enable eviction of the tail frames to free up storage to put the new frame in on OOM. Currently, we return an error on put frame.

MushMal:
Done

PR #24, opened 2020-03-18
Adding dropping tail frames policy on the storage OOM

MushMal:
Issue #20  https://github.com/awslabs/amazon-kinesis-video-streams-pic/issues/22

Exposing policy for content view to trim the tail until next fragment start
Exposing policy for the content store to not OOM but trim tail
Reving the stream info structure with backward compat
Renamed the existing enum for the content view policy to match/look similar to the content store policy.
Added unit tests


Sean-Der (approved):

src/client/src/Stream.c
```diff
@@ -1598,14 +1600,22 @@ STATUS waitForAvailability(PKinesisVideoStream pKinesisVideoStream, UINT32 alloc
...
+        if (IS_OFFLINE_STREAMING_MODE(pKinesisVideoStream->streamInfo.streamCaps.streamingType)) {
+            // Long path which will await for the availability notification or cancellation
+            CHK_STATUS(pKinesisVideoClient->clientCallbacks.waitConditionVariableFn(
+                    pKinesisVideoClient->clientCallbacks.customData,
+                    pKinesisVideoStream->bufferAvailabilityCondition,
+                    pKinesisVideoStream->base.lock,
+                    pKinesisVideoClient->deviceInfo.clientInfo.offlineBufferAvailabilityTimeout));
+        } else {
+            // Need to evict the tail frames by trimming the tail
+            CHK_STATUS(contentViewGetTail(pKinesisVideoStream->pView, &pViewItem));
```
chehefen: this behavior should also be affected by viewOverflowPolicy (drop one item vs drop fragment.)

MushMal: Yes indeed. For some reason I entirely forgot the reason I rebased this on top of your change :) Will send an update

chehefen (approved):

PR #37, opened 2020-03-29
[Issue-36] Add logic in thread sleep to handle large sleep values

### LINKED ISSUE #36: Modify the thread sleep logic to work for values greater than useconds_t
Currently, the thread sleep function calls usleep with the argument being passed to it. However, this causes very large sleep values to not behave as expected.

disa6302:
*Issue #, if available: #36 *

*Description of changes:*
- Added logic in THREAD_SLEEP to cater to cases where the sleep time values would be more than the maximum value supported by usleep() / Sleep() (0xffffffff)
- Tested locally with different sleep values for if - else coverage

Resolves #36 

### DIFF
--- a/src/utils/src/Thread.c
+++ b/src/utils/src/Thread.c
@@ -78,7 +78,23 @@ CleanUp:
 
 PUBLIC_API VOID defaultThreadSleep(UINT64 time)
 {
-    Sleep((UINT32) (time / HUNDREDS_OF_NANOS_IN_A_MILLISECOND));
+    // Time in milliseconds
+    UINT64 remaining_time = time / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+    // The loop will be run till the complete Sleep() time is reached
+    while(remaining_time != 0) {
+        // Covers the last case when there is residual time left and
+        // when the value provided is less than or equal to MAX_UINT32
+        if(remaining_time <= MAX_UINT32) {
+            Sleep((UINT32) remaining_time);
+            remaining_time = 0;
+        }
+        // Sleep maximum time supported by Sleep() repeatedly to cover large
+        // sleep time cases
+        else {
+            Sleep(MAX_UINT32);
+            remaining_time = remaining_time - (UINT64) MAX_UINT32;
+        }
+    }
 }
 
 PUBLIC_API STATUS defaultCancelThread(TID threadId)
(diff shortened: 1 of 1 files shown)


src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -364,6 +364,8 @@ typedef CID*                PCID;
 #define MAX_INT64           ((INT64) 0x7fffffffffffffff)
 #define MIN_INT64           ((INT64) 0x8000000000000000)
```
MushMal: Don’t really need this. You could use MAX_UINT32 instead

src/utils/src/Thread.c
```diff
@@ -78,7 +78,23 @@ PUBLIC_API STATUS defaultJoinThread(TID threadId, PVOID* retVal)
 
 PUBLIC_API VOID defaultThreadSleep(UINT64 time)
 {
-    Sleep((UINT32) (time / HUNDREDS_OF_NANOS_IN_A_MILLISECOND));
+    // Time in milliseconds
+    UINT64 remaining_time = time / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+    // The loop will be run till the complete Sleep() time is reached
+    while(remaining_time > 0) {
```
MushMal: For unsigned integers it’s more prudent to check against not being 0 only

src/utils/src/Thread.c
```diff
@@ -300,4 +334,4 @@ threadSleep globalThreadSleep = defaultThreadSleep;
 threadSleepUntil globalThreadSleepUntil = defaultThreadSleepUntil;
 joinThread globalJoinThread = defaultJoinThread;
 cancelThread globalCancelThread = defaultCancelThread;
-detachThread globalDetachThread = defaultDetachThread;
\ No newline at end of file
+detachThread globalDetachThread = defaultDetachThread;
```
MushMal: Please leave the \n and the end of the file

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
disa6302: Removed

src/utils/src/Thread.c
disa6302: Done

Issue #26, opened 2020-03-18
Implement graceful handling for intermittent producer in PIC


MushMal:
If the producer media pipeline stops producing frames, the backend will timeout which will cause the automatic re-connect in case of re-streaming flag is specified (default). We should check if the buffer is free and whether we received the last ACK in case of persistence is enabled and move the state machine to ready state only.

PR #40, opened 2020-04-02
Implementing intermittent producer support in the PIC

MushMal:
Issue #26 

Adding logic to PIC to not auto-prime on timeout upload handle termination if we have no more data in the buffer to be streamed out.

### DIFF
--- a/src/client/src/StreamState.c
+++ b/src/client/src/StreamState.c
@@ -705,6 +705,7 @@ STATUS executeStoppedStreamState(UINT64 customData, UINT64 time)
     ENTERS();
     UNUSED_PARAM(time);
     STATUS retStatus = STATUS_SUCCESS;
+    UINT64 duration, viewByteSize;
     PKinesisVideoStream pKinesisVideoStream = STREAM_FROM_CUSTOM_DATA(customData);
 
     CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
@@ -712,7 +713,26 @@ STATUS executeStoppedStreamState(UINT64 customData, UINT64 time)
     // Also store the connection stopping result
     pKinesisVideoStream->connectionDroppedResult = pKinesisVideoStream->base.result;
 
-    // Step the client state machine first
+    // Check if we want to prime the state machine based on whether we have any more content to send
+    // currently and if the error is a timeout.
+    if (SERVICE_CALL_RESULT_IS_A_TIMEOUT(pKinesisVideoStream->connectionDroppedResult)) {
+        CHK_STATUS(getAvailableViewSize(pKinesisVideoStream, &duration, &viewByteSize));
+        if (viewByteSize == 0) {
+            // Next time putFrame is called we will self-prime.
+            // The idea is to drop frames till new key frame which will
+            // become a stream start.
+            pKinesisVideoStream->resetGeneratorOnKeyFrame = TRUE;
+            pKinesisVideoStream->skipNonKeyFrames = TRUE;
+
+            // Set an indicator to step the state machine on new frame
+            pKinesisVideoStream->streamState = STREAM_STATE_NEW;
+
+            // Early exit
+            CHK(FALSE, retStatus);
+        }
+    }
+
+    // Auto-prime the state machine
     CHK_STATUS(stepStateMachine(pKinesisVideoStream->base.pStateMachine));
 
 CleanUp:
--- a/src/client/src/StreamState.c
+++ b/src/client/src/StreamState.c
@@ -265,48 +265,69 @@ STATUS fromPutStreamState(UINT64 customData, PUINT64 pState) {
     // Transition to states if not stopped
     if (pKinesisVideoStream->streamState == STREAM_STATE_STOPPED) {
         state = STREAM_STATE_STOPPED;
-    } else if (pKinesisVideoStream->base.result == SERVICE_CALL_RESULT_OK) {
-        // find the handle in the new state and set it to ready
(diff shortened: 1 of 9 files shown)


src/client/src/StreamState.c
```diff
@@ -705,14 +705,34 @@ STATUS executeStoppedStreamState(UINT64 customData, UINT64 time)
...
 
     CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
 
     // Also store the connection stopping result
     pKinesisVideoStream->connectionDroppedResult = pKinesisVideoStream->base.result;
 
-    // Step the client state machine first
+    // Check if we want to prime the state machine based on whether we have any more content to send
+    // currently and if the error is a timeout.
+    if (SERVICE_CALL_RESULT_IS_A_TIMEOUT(pKinesisVideoStream->connectionDroppedResult)) {
```
chehefen, line 718: Did you consider putting this logic inside streamTerminatedEvent? I think any kind of timeout would have to go through that path. In there we can simply do if viewByteSize == 0 then resetStream.

MushMal: streamTerminatedEvent is only applicable for the upload handles - i,e, when we are streaming already. It actually does all the right things. The differentiation is to not auto-prime the state machine which is what we are doing here.

MushMal: Also, resetStream is a very tough measure with dumping the frame buffers. It's also kind of orthogonal - the main idea here is to do "idling"

src/client/src/StreamState.c
```diff
@@ -705,14 +705,34 @@ STATUS executeStoppedStreamState(UINT64 customData, UINT64 time)
...
+        CHK_STATUS(getAvailableViewSize(pKinesisVideoStream, &duration, &viewByteSize));
+        if (viewByteSize == 0) {
+            // Next time putFrame is called we will self-prime.
+            // The idea is to drop frames till new key frame which will
+            // become a stream start.
+            pKinesisVideoStream->resetGeneratorOnKeyFrame = TRUE;
+            pKinesisVideoStream->skipNonKeyFrames = TRUE;
+
+            // Set an indicator to step the state machine on new frame
+            pKinesisVideoStream->streamState = STREAM_STATE_NEW;
```
chehefen, line 728: if user has been streaming and then stopped, shouldnt it go to putStream state? when user start streaming again, it just open putMedia straight away.

src/client/src/StreamState.c
```diff
@@ -705,14 +705,34 @@ STATUS executeStoppedStreamState(UINT64 customData, UINT64 time)
...
-    // Step the client state machine first
+    // Check if we want to prime the state machine based on whether we have any more content to send
+    // currently and if the error is a timeout.
+    if (SERVICE_CALL_RESULT_IS_A_TIMEOUT(pKinesisVideoStream->connectionDroppedResult)) {
+        CHK_STATUS(getAvailableViewSize(pKinesisVideoStream, &duration, &viewByteSize));
+        if (viewByteSize == 0) {
+            // Next time putFrame is called we will self-prime.
+            // The idea is to drop frames till new key frame which will
+            // become a stream start.
+            pKinesisVideoStream->resetGeneratorOnKeyFrame = TRUE;
```
chehefen, line 724: is this gonna cause two mkv headers to be generated? we also generate header in streamStartFixupOnReconnect.

MushMal: A single header. The changes in the latest test hits this case to validate.

src/client/src/StreamState.c
MushMal: Excellent question. Right now, on a timeout it would go to STREAM_STATE_GET_ENDPOINT state. The existing code assumes that if the endpoint timed out then perhaps it's the best to get the new one. However, in case of intermittent producer, perhaps it makes sense to go directly to STREAM_STATE_PUT_STREAM state and if that fails with a timeout then go to the endpoint state. Let me make adjustments

chehefen:
can you check out this branch in cproducer and make sure the cproducer tests pass too?

MushMal:
> can you check out this branch in cproducer and make sure the cproducer tests pass too?

There are few failures in the test but those will be fixed in their due course. I need to ensure that there is no major regression which I am doing right now. We will update C Producer to this commit once we fix the tests and ensure everything works OK. 

This said, the logic in these changes looks to be right

chehefen:
> > can you check out this branch in cproducer and make sure the cproducer tests pass too?
> 
> There are few failures in the test but those will be fixed in their due course. I need to ensure that there is no major regression which I am doing right now. We will update C Producer to this commit once we fix the tests and ensure everything works OK.
> 
> This said, the logic in these changes looks to be right

If cproducer test failed that doesnt count as regression?

MushMal:
> > > can you check out this branch in cproducer and make sure the cproducer tests pass too?
> > 
> > 
> > There are few failures in the test but those will be fixed in their due course. I need to ensure that there is no major regression which I am doing right now. We will update C Producer to this commit once we fix the tests and ensure everything works OK.
> > This said, the logic in these changes looks to be right
> 
> If cproducer test failed that doesnt count as regression?

As is it is. The tests in the C producer are geared towards the existing behavior. We are changing the behavior for the intermittent producer. Those are the tests that are failing with the new PIC.

I am looking at the failures locally and vetting them off with the new logic. So far, nothing that doesn't explicitly deal with the intermittent scenario broken.

Once we rebase the PIC in C producer, we need to fix the intermittent tests.

PR #41, opened 2020-04-08
Adding instrumented allocators to util

MushMal:
Having an optional compile-time set/reset macros
Minor warning fixes
Unit tests

NOTE: Many of the unit tests overwrite the allocators to do final memory leak tracking. These are Client, Util, C producer, WebRTC Signaling Client unit tests. The instrumented allocator exposed publicly will be useful for the client application debugging any memory leaks. 

The main pattern is to build the application which would use the conditional compilation macros in the commit which in normal compile mode would be no-ops. In case of INSTRUMENTED_ALLOCATORS being set at compile time, the macros will evaluate to the actual calls to the APIs to set the instrumented allocators and re-set them. The pattern of usage will be - when the application starts, it calls SET_INSTRUMENTED_ALLOCATORS() before any allocations are made and on exit it will call RESET_INSTRUMENTED_ALLOCATORS() after all of the allocations are freed. The printout or the return value from the reset will indicate whether there is any memory leaks. 

Another pattern is for the canaries/long-run stress tests. The instrumented allocator not only allows for final checks for any leaks but also periodic reporting of total memory usage metric.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/utils/src/Allocators.c
+++ b/src/utils/src/Allocators.c
@@ -65,6 +65,9 @@ VOID dumpMemoryHex(PVOID pMem, UINT32 size)
     DLOGS("++++++++++++++++++++++++++++++++++++++++++++");
     DLOGS("Dumping memory done!");
     DLOGS("============================================");
+#else
+    UNUSED_PARAM(pMem);
+    UNUSED_PARAM(size);
 #endif
 }
 
--- a/src/utils/src/InstrumentedAllocators.c
+++ b/src/utils/src/InstrumentedAllocators.c
@@ -1,0 +1,148 @@
+#include "Include_i.h"
+
+volatile SIZE_T gInstrumentedAllocatorsTotalAllocationSize = 0;
+
+memAlloc gInstrumentedAllocatorsStoredMemAlloc;
+memAlignAlloc gInstrumentedAllocatorsStoredMemAlignAlloc;
+memCalloc gInstrumentedAllocatorsStoredMemCalloc;
+memFree gInstrumentedAllocatorsStoredMemFree;
+memRealloc gInstrumentedAllocatorsStoredMemRealloc;
+
+STATUS setInstrumentedAllocators()
+{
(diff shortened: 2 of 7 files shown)


src/utils/src/Allocators.c
```diff
@@ -86,4 +89,4 @@ BOOL checkBufferValues(PVOID ptr, BYTE val, SIZE_T size)
     return TRUE;
 }
 
-memChk globalMemChk = checkBufferValues;
\ No newline at end of file
+memChk globalMemChk = checkBufferValues;
```
disa6302, line 92: New line after this.

src/utils/src/InstrumentedAllocators.c
```diff
@@ -0,0 +1,148 @@
...
+    // Add to the total book keeping
+    ATOMIC_ADD(&gInstrumentedAllocatorsTotalAllocationSize, size);
+
+    return pAlloc + 1;
+}
+
+PVOID instrumentedAllocatorsMemAlignAlloc(SIZE_T size, SIZE_T alignment)
+{
+    // Doing the same as alloc
+    DLOGS("Instrumented mem alignalloc %" PRIu64 " bytes", (UINT64) size);
```
disa6302, line 83: Why arent we supporting this right now?

src/utils/src/Allocators.c
MushMal: I believe it's the diffing tool issue - the code has the trailing \n at the bottom of the file

src/utils/src/InstrumentedAllocators.c
MushMal: The main reason is that in the KVS Client we have a flag that sets up the content store as the allocator which will be tightly packed by default. In this case, the align alloc will be meaningless. Also, we don't use this anywhere in the codebase. I think the best would be to simply remove this API at some stage.

PR #42, opened 2020-04-09
replace CHK_LOG_ERR_NV with CHK_LOG_ERR. Put CHK_LOG_ERR at the exit …

chehefen:
…of each interface function

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/client/src/Client.c
+++ b/src/client/src/Client.c
@@ -167,6 +167,8 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
 
 CleanUp:
 
+    CHK_LOG_ERR(retStatus);
+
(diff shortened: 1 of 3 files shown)


src/client/src/Client.c
```diff
@@ -664,6 +677,8 @@ STATUS putKinesisVideoFrame(STREAM_HANDLE streamHandle, PFrame pFrame)
         DLOGW("Failed to submit frame to Kinesis Video client. "
                       "status: 0x%08x decoding timestamp: %" PRIu64 " presentation timestamp: %" PRIu64,
               retStatus, pFrame->decodingTs, pFrame->presentationTs);
+    } else {
```
MushMal, line 680: Not sure why you need this here

chehefen: in case frame is null for some reason.

Issue #46, opened 2020-04-18
Enhanced stream and client metrics


MushMal:
Implement extra metrics for bot the client and the stream

MushMal:
These will be very useful for various diagnostics purposes as well as for specialized tools calculating latencies and reporting those.

disa6302:
Closing this issue.

PR #47, opened 2020-04-18
Enhanced stream and client metrics

MushMal:
Issue #46 

Initial work for enhanced stream level metrics. More to follow. Would like to iterate fast on these so this is more of a private PR.

### DIFF
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -1182,6 +1182,9 @@ struct __StreamMetrics {
     // Total number of skipped frames
     UINT64 skippedFrames;
 
+    // Total number of storage pressure events
+    UINT64 storagePressures;
+
     // Total number of latency pressure events
     UINT64 latencyPressures;
 
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -448,7 +448,7 @@ extern "C" {
 #define SERVICE_CALL_CONTEXT_CURRENT_VERSION                0
 #define STREAM_DESCRIPTION_CURRENT_VERSION                  1
 #define FRAGMENT_ACK_CURRENT_VERSION                        0
-#define STREAM_METRICS_CURRENT_VERSION                      0
+#define STREAM_METRICS_CURRENT_VERSION                      1
 #define CLIENT_METRICS_CURRENT_VERSION                      0
 #define CLIENT_INFO_CURRENT_VERSION                         0
 
@@ -1123,6 +1123,8 @@ struct __StreamMetrics {
     // Version of the struct
     UINT32 version;
 
+    // V0 metrics following
+
     // Duration from the current to the head in 100ns
     UINT64 currentViewDuration;
 
@@ -1140,6 +1142,60 @@ struct __StreamMetrics {
 
     // Last measured transfer rate in bytes per second
     UINT64 currentTransferRate;
+
+    // V1 metrics following
+
+    // Total stream duration - stream uptime
+    UINT64 uptime;
+
+    // Total transferred bytes
+    UINT64 transferredBytes;
+
+    // Total number of streaming sessions including new/errored/active
+    UINT64 totalSessions;
+
+    // Total number of active streaming sessions - only the ones that actually have streamed some data
+    UINT64 totalActiveSessions;
+
+    // Total number of buffered ACKs
+    UINT64 bufferedAcks;
+
+    // Total number of received ACKs
+    UINT64 receivedAcks;
+
+    // Total number of persisted ACKs
+    UINT64 persistedAcks;
+
+    // Total number of error ACKs
+    UINT64 errorAcks;
+
+    // Total number of dropped frames
+    UINT64 droppedFrames;
+
+    // Total number of dropped fragments
+    // TODO: Near-realtime streaming mode is not used currently supported and this value will be 0
+    UINT64 droppedFragments;
+
+    // Total number of skipped frames
+    UINT64 skippedFrames;
+
+    // Total number of latency pressure events
+    UINT64 latencyPressures;
+
+    // Total number of Buffer pressure events
+    UINT64 bufferPressures;
+
+    // Total number of stream stale events
+    UINT64 staleEvents;
+
+    // Total number of put frame call errors
+    UINT64 putFrameErrors;
+
+    // Backend Control Plane API call latency which includes success and failure
+    UINT64 cplApiCallLatency;
+
+    // Backend Data Plane API call latency which includes success and failure
+    UINT64 dataApiCallLatency;
 };
 
 typedef struct __StreamMetrics* PStreamMetrics;
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -450,7 +450,7 @@ extern "C" {
 #define FRAGMENT_ACK_CURRENT_VERSION                        0
 #define STREAM_METRICS_CURRENT_VERSION                      1
 #define CLIENT_METRICS_CURRENT_VERSION                      0
-#define CLIENT_INFO_CURRENT_VERSION                         0
+#define CLIENT_INFO_CURRENT_VERSION                         1
 
 /**
  * Definition of the client handle
@@ -1052,6 +1052,11 @@ typedef struct __ClientInfo {
 
     // whether to log metric or not
     BOOL logMetric;
+
+    // ------------------------------- V0 compat ----------------------
+
+    // Time that allowed to be elapsed between the metric loggings if enabled
+    UINT64 metricLoggingPeriod;
 } ClientInfo, *PClientInfo;
 
 /**
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -1157,6 +1157,9 @@ struct __StreamMetrics {
     // Total number of active streaming sessions - only the ones that actually have streamed some data
     UINT64 totalActiveSessions;
 
+    // Average session duration
+    UINT64 avgSessionDuration;
+
     // Total number of buffered ACKs
     UINT64 bufferedAcks;
 
--- a/src/client/src/Client.c
+++ b/src/client/src/Client.c
@@ -1392,6 +1392,8 @@ VOID viewItemRemoved(PContentView pContentView, UINT64 customData, PViewItem pVi
         switch (pKinesisVideoStream->streamInfo.streamCaps.streamingType) {
             case STREAMING_TYPE_REALTIME:
             case STREAMING_TYPE_OFFLINE:
+                pKinesisVideoStream->diagnostics.droppedFrames++;
+
                 // The callback is optional so check if specified first
                 if (pKinesisVideoClient->clientCallbacks.droppedFrameReportFn != NULL) {
                     CHK_STATUS(pKinesisVideoClient->clientCallbacks.droppedFrameReportFn(
@@ -1403,6 +1405,8 @@ VOID viewItemRemoved(PContentView pContentView, UINT64 customData, PViewItem pVi
                 break;
 
             case STREAMING_TYPE_NEAR_REALTIME:
+                pKinesisVideoStream->diagnostics.droppedFragments++;
+
                 // The callback is optional so check if specified first
                 if (pKinesisVideoClient->clientCallbacks.droppedFragmentReportFn != NULL) {
                     CHK_STATUS(pKinesisVideoClient->clientCallbacks.droppedFragmentReportFn(
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -284,6 +284,9 @@ STATUS createStream(PKinesisVideoClient pKinesisVideoClient, PStreamInfo pStream
     // Set the initial state to new
     pKinesisVideoStream->streamState = STREAM_STATE_NEW;
 
+    // Store the stream uptime start
+    pKinesisVideoStream->diagnostics.createTime = pKinesisVideoClient->clientCallbacks.getCurrentTimeFn(pKinesisVideoClient->clientCallbacks.customData);
+
     // Call to transition the state machine
     CHK_STATUS(stepStateMachine(pKinesisVideoStream->base.pStateMachine));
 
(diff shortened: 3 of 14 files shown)


src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+    UINT64 uptime;
+
+    // Total transferred bytes
+    UINT64 transferredBytes;
+
+    // Total number of streaming sessions including new/errored/active
+    UINT64 totalSessions;
+
+    // Total number of active streaming sessions - only the ones that actually have streamed some data
+    UINT64 totalActiveSessions;
```
chehefen, line 1158: do we need this? this should be at most 2.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+    UINT64 droppedFragments;
+
+    // Total number of skipped frames
+    UINT64 skippedFrames;
+
+    // Total number of latency pressure events
+    UINT64 latencyPressures;
+
+    // Total number of Buffer pressure events
+    UINT64 bufferPressures;
```
chehefen, line 1189: what is bufferPressure? is it storagePressure?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+    // V1 metrics following
+
+    // Total stream duration - stream uptime
+    UINT64 uptime;
+
+    // Total transferred bytes
+    UINT64 transferredBytes;
+
+    // Total number of streaming sessions including new/errored/active
+    UINT64 totalSessions;
```
disa6302, line 1155: I am wondering what the use case for total sessions would be?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
     // Last measured transfer rate in bytes per second
     UINT64 currentTransferRate;
+
+    // V1 metrics following
+
+    // Total stream duration - stream uptime
+    UINT64 uptime;
+
+    // Total transferred bytes
+    UINT64 transferredBytes;
```
disa6302, line 1152: How often would we be re-calculating this? Would it be at every key frame?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+
+    // Total number of dropped fragments
+    // TODO: Near-realtime streaming mode is not used currently supported and this value will be 0
+    UINT64 droppedFragments;
+
+    // Total number of skipped frames
+    UINT64 skippedFrames;
+
+    // Total number of latency pressure events
+    UINT64 latencyPressures;
```
chehefen, line 1186: latency pressure is emitted on every frame, is this metric gonna be useful?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+
+    // Total number of put frame call errors
+    UINT64 putFrameErrors;
+
+    // Backend Control Plane API call latency which includes success and failure
+    UINT64 cplApiCallLatency;
+
+    // Backend Data Plane API call latency which includes success and failure
+    UINT64 dataApiCallLatency;
 };
```
disa6302, line 1202: Just throwing it out here. What do you think about having metrics tracking different retries? I am unsure of the different retries within the stack though.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1142,63 @@ struct __StreamMetrics {
...
+    UINT64 latencyPressures;
+
+    // Total number of Buffer pressure events
+    UINT64 bufferPressures;
+
+    // Total number of stream stale events
+    UINT64 staleEvents;
+
+    // Total number of put frame call errors
+    UINT64 putFrameErrors;
```
chehefen, line 1195: if this is the total number of putFrameErrors, how do we detect that it recovers? The monitoring system then has to keep state of what the previous putFrameError count was.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: For devices perhaps. For NVR it could be hundreds

MushMal: Sorry, this is the sessions. This is an indicator of how many sessions had been active. Not the current ones.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: Buffer pressure

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: Latency pressure will be emitted ONLY when we cross the MAX_LATENCY threshold. This is very similar to buffer pressure - just triggered on a different threshold

MushMal:
I will be adding to client metrics very soon. Let me do that in this CR instead of spinning another one

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: Yes, this is only a report from the low-level. We need a separate non-recoverable error. It is not going to be easy due to the nature of error handling and the different configurations.

chehefen:
I think before this go in we should have a doc written up going over what each metric mean and how they can be monitored. Then it should be reviewed by the team before it is committed into code.

MushMal:
> I think before this go in we should have a doc written up going over what each metric mean and how they can be monitored. Then it should be reviewed by the team before it is committed into code.

This is exactly what had been done. The documentation though is something that needs to follow-through. So far, the documentation is on the public API in Include.h. Please comment whether something doesn't make sense. We can take this offline to discuss further.

src/client/src/Stream.h
```diff
@@ -102,11 +102,66 @@ struct __KinesisVideoStreamDiagnostics {
...
+
+    // Total number of put frame call errors
+    UINT64 putFrameErrors;
+
+    // Backend Control Plane API call latency which includes success and failure
+    UINT64 cplApiCallLatency;
+
+    // Backend Data Plane API call latency which includes success and failure
+    UINT64 dataApiCallLatency;
 };
```
disa6302, line 165: One more metric I can think of is number of recoverable errors?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -448,7 +448,7 @@ extern "C" {
 #define SERVICE_CALL_CONTEXT_CURRENT_VERSION                0
 #define STREAM_DESCRIPTION_CURRENT_VERSION                  1
 #define FRAGMENT_ACK_CURRENT_VERSION                        0
-#define STREAM_METRICS_CURRENT_VERSION                      0
+#define STREAM_METRICS_CURRENT_VERSION                      1
 #define CLIENT_METRICS_CURRENT_VERSION                      0
```
disa6302, line 452: No need to update client metrics version?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: Not all of the metrics would be used with the same frequency or need. The total number of session is useful to compare against the total active session number to monitor whether we have sessions that are not streaming for some reason. In general, those numbers should be identical.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: With every filled chunk

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: At this locality, we can't track the retriable/non-retriable errors for example as those depend on a number of factors including the scenario.

We should have separate metrics tracking fatal errors perhaps but those should be tracked at a higher layer.

For example. A stream could be configured to not retry. In this case a retriable error will be fatal. Another example is an auth integration - this might be retriable or fatal - depending on the integration. Similar to this is the network errors or media errors.

I believe that we can't fix every possible situation. These are the basic metrics needed. We will have to add more heuristics/higher-logic dependent metrics as an iterative process.

src/client/src/Stream.h
MushMal: As I mentioned, it's hard to tell which one is recoverable and not. We can add more heuristics/scenario driven recoverable/fatal errors metric later

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: no - haven't changed that structure

src/client/include/com/amazonaws/kinesis/video/client/Include.h
disa6302: Understood. Makes sense.

src/client/src/Client.c
```diff
@@ -1403,6 +1405,8 @@ VOID viewItemRemoved(PContentView pContentView, UINT64 customData, PViewItem pVi
                 break;
 
             case STREAMING_TYPE_NEAR_REALTIME:
+                pKinesisVideoStream->diagnostics.droppedFragments++;
```
disa6302, line 1408: We drop fragments in near realtime, but frames in offline mode? Can you explain why the difference between the 2 streaming modes?

src/client/src/Stream.c
```diff
@@ -659,31 +662,40 @@ STATUS logStreamMetric(PKinesisVideoStream pKinesisVideoStream)
...
+    DLOGD("\tTotal streams frame rate (fps): %" PRIu64 " ", clientMetrics.totalFrameRate);
+    DLOGD("\tTotal streams transfer rate (bps): %" PRIu64 " (%" PRIu64 " Kbps)", clientMetrics.totalTransferRate * 8, clientMetrics.totalTransferRate * 8 / 1024);
+    DLOGD("\tCurrent view duration (ms): %" PRIu64 " ", streamMetrics.currentViewDuration / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    DLOGD("\tOverall view duration (ms): %" PRIu64 " ", streamMetrics.overallViewDuration / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    DLOGD("\tCurrent view byte size: %" PRIu64 " ", streamMetrics.currentViewSize);
+    DLOGD("\tOverall view byte size: %" PRIu64 " ", streamMetrics.overallViewSize);
+    DLOGD("\tCurrent frame rate (fps): %f ", streamMetrics.currentFrameRate);
+    DLOGD("\tCurrent transfer rate (bps): %" PRIu64 " (%" PRIu64 " Kbps)", streamMetrics.currentTransferRate * 8, streamMetrics.currentTransferRate * 8 / 1024);
+
+    // V1 information
```
disa6302, line 679: I understand we need to report the metrics we calculate, but, considering the current rate at which the information is logged (at every key frame), I feel this will be a lot of information in the logs. Should we display all the V1 metrics in the logs?

sharadrj: I share the concern. Would be better if we print the metric once per minute or so If it is not easily doable from PIC, better to not print anything at this layer and leave all printing to C-Producer.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
sharadrj: Lol. I would like to understand this point also.

src/client/src/Client.c
MushMal: The SDK defines 3 modes of streaming

* Realtime - this is when a camera-like source produces frames which get pushed into the SDK and streamed - the main scenario.
* Offline - this is when frames are produced in IO/CPU speed into the SDK. This is a file upload like scenario. The main difference between this and Realtime mode is that this will block on "availability" (temporal and storage) while the the Realtime mode will push tail frames out. There are other nuances too in the way we process ACKs.
* Near-realtime - this mode is not implemented yet but this will operate at a fragment granularity. Similar to HLS/MPEG-DASH but for upload. The fragment(s) will be assembled and pushed in one go in a single session.

src/client/src/Stream.c
MushMal: I also share this concern. Especially, since this is in PIC and especially since the produces could have multiple streams. Especially, since those streams could be audio when every frame is a key-frame and some are being produced at hundreds of FPS.

We do printout in the CPP layer too.

I would like to address this properly. Let me add some changes for handling this. I think the best would be to log on a key-frame after certain time elapses controlled by a pass-in variable. Currently, it's gated on a boolean pKinesisVideoClient->deviceInfo.clientInfo.logMetric but I will make it a period

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1147,66 @@ struct __StreamMetrics {
...
+
+    // Total number of dropped frames
+    UINT64 droppedFrames;
+
+    // Total number of dropped fragments
+    // TODO: Near-realtime streaming mode is not used currently supported and this value will be 0
+    UINT64 droppedFragments;
+
+    // Total number of skipped frames
+    UINT64 skippedFrames;
```
disa6302, line 1188: How would this be useful for the application layer? This would mostly be useful for statistical analysis, right? Just wondering how the application could benefit from skipped frames metrics to perform any action based on it.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1052,6 +1052,11 @@ typedef struct __ClientInfo {
 
     // whether to log metric or not
     BOOL logMetric;
+
+    // ------------------------------- V0 compat ----------------------
+
+    // Time that allowed to be elapsed between the metric loggings if enabled
+    UINT64 metricLoggingPeriod;
```
zhiyua-git, line 1059: If we are going to do c producer level sampling and aggregation, why does PIC need to hold metricLoggingPeriod?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
zhiyua-git: pressure on buffer duration?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1140,6 +1147,66 @@ struct __StreamMetrics {
...
+    UINT64 staleEvents;
+
+    // Total number of put frame call errors
+    UINT64 putFrameErrors;
+
+    // Backend Control Plane API call latency which includes success and failure
+    UINT64 cplApiCallLatency;
+
+    // Backend Data Plane API call latency which includes success and failure
+    UINT64 dataApiCallLatency;
```
zhiyua-git, line 1209: Shouldn't this and control plane api latency only track the successful ones? There should be a separate failure/timeout metrics IMO.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: I agree, mostly for analysis. The application can have some general idea how many frames are dropped/skipped by looking at the statistics over some longer period of time.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: This really indicates the amount of time that should elapse for the DLOGD printout. Useful for debug logs. Previously, we were logging debug log on every key frame which is terrible if we are handling audio stream only with all key frames and hundreds of FPS.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
MushMal: WOW - just noticed that I added a common API to calculate the latency but only invoking it in desctibe call. Let me update it. I also think that we should handle the erred cases too - for example the timeout.

RE: bufferPressures - this is tracking the number of times the buffer pressure events have been invoked. This is slightly different callback than latency pressure in sense that it looks at the overall buffer duration rather than MAX_LATENCY.

PR #54, opened 2020-05-12
Enhance build system for users

lherman-cs:
Changes:
* Add install steps
* Add pc files

### DIFF
--- a/CMake/Utilities.cmake
+++ b/CMake/Utilities.cmake
@@ -36,30 +36,30 @@ function(build_dependency lib_name)
   # library building cmake.
   set(build_args ${ARGN})
 
-  file(REMOVE_RECURSE ${KINESIS_VIDEO_OPEN_SOURCE_SRC}/lib${lib_name})
+  file(REMOVE_RECURSE ${OPEN_SRC_INSTALL_PREFIX}/lib${lib_name})
 
   # build library
   configure_file(
-    ${CMAKE_SOURCE_DIR}/CMake/Dependencies/lib${lib_name}-CMakeLists.txt
-    ${KINESIS_VIDEO_OPEN_SOURCE_SRC}/lib${lib_name}/CMakeLists.txt COPYONLY)
+    ./CMake/Dependencies/lib${lib_name}-CMakeLists.txt
+    ${OPEN_SRC_INSTALL_PREFIX}/lib${lib_name}/CMakeLists.txt COPYONLY)
   execute_process(
     COMMAND ${CMAKE_COMMAND} ${build_args}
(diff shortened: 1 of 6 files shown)


CMake/Utilities.cmake
```diff
@@ -36,30 +36,30 @@ function(build_dependency lib_name)
...
   set(build_args ${ARGN})
 
-  file(REMOVE_RECURSE ${KINESIS_VIDEO_OPEN_SOURCE_SRC}/lib${lib_name})
+  file(REMOVE_RECURSE ${OPEN_SRC_INSTALL_PREFIX}/lib${lib_name})
 
   # build library
   configure_file(
-    ${CMAKE_SOURCE_DIR}/CMake/Dependencies/lib${lib_name}-CMakeLists.txt
-    ${KINESIS_VIDEO_OPEN_SOURCE_SRC}/lib${lib_name}/CMakeLists.txt COPYONLY)
+    ./CMake/Dependencies/lib${lib_name}-CMakeLists.txt
```
MushMal, line 43: Does cmake have a pre-configured variable for this path?

CMakeLists.txt
```diff
@@ -112,19 +113,49 @@ if(UNIX AND NOT APPLE)
...
+
 add_library(kvspicUtils STATIC ${PIC_UTILS_SOURCE_FILES})
 target_link_libraries(kvspicUtils ${CMAKE_DL_LIBS} Threads::Threads)
 if(UNIX AND NOT APPLE)
   # rt needed for clock_gettime
   target_link_libraries(kvspicUtils rt)
 endif()
 
+install(
+  TARGETS kvspic kvspicClient kvspicState kvspicUtils
```
MushMal, line 142: Will this properly create a directory per project for the include files? I assume so as we would end up with name conflicts as all the public includes are named Include.h

libkvspic.pc.cmake
```diff
@@ -0,0 +1,10 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=${prefix}
+includedir=${prefix}/include
```
MushMal, line 3: Oh, there it is

CMake/Utilities.cmake
lherman-cs: AFAIK, this path is custom to us, so I don't think CMAKE have a pre-configure variable for this.

CMakeLists.txt
lherman-cs: Yep, the prefix is configurable using `CMAKE_INSTALL_PREFIX`

Issue #43, opened 2020-04-15
Skip SEI NALu in getVideoWidthAndHeightFromH264SPS


MushMal:
When processing H264 (similar issue with H265 potentially) we are extracting the width and height of the picture from the SPS in order to set MKV parameters. This is for Video H264 (and H265) only. The function assumes that the first NALu of the CPD is SPS whereas it might be SEI.

The fix is to check the first NALu and skip it if it's SEI and process SPS only.

farleylai:
Hi @MushMal,

It seems like the SEI and multiple SPS/PPS in the key frame are handled.
I have a source giving key frames with NALUs as follows:
- key frame with SPS/PPS/SEI but no IDR followed by
- key frame with SPS/PPS/SPS/PPS/IDR

KVS SDK and the console do not return or show any errors this time.
Can you confirm that is the case?

MushMal:
@farleylai the SDK attempts to validate only stuff that's absolutely necessary. The SDK passes along the data that it validates at the packaging layer to the backend. The similar strategy is on the backend, we try to validate only items that are needed for streamability, indexing and most importantly, correctness. So, multiple properly formed SPS/PPS/SEI should be OK from the packaging standpoint but you might also want to ensure that the player/playback solution you use downstream are OK.
The issue I opened is about trying to extract the width and height to set MKV video element width and height - these really are useless but some downstream players seem to care. Similar situation is for Audio content.

farleylai:
@MushMal 
Sorry, I intended to post in the original issue, not here.
And agreed, this is a separate concern which is not in our use cases yet.

MushMal:
Now it should work -> https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/68

I reverted the removal of the non-VCL NALu changes. We had an issue with an extra frame size bloat due to the extra zeroes

MushMal:
Changes are in

PR #68, opened 2020-06-18
Handling bad encoders producing extra 0 at the end of the NALu run. H…

MushMal:
…andling AUD and SEI NALus. Adding tests.

Issue #43 
Issue #https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/491

*Description of changes:*


chehefen:
I dont think you need to remove the non-VCL frames NALus. I used gstreamer to do the nal adaption in issue https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/491 and it plays fine in the console. I checked the frame that gstreamer generated and it preserves all NALus in the original frame.

src/mkvgen/src/NalAdapter.c
```diff
@@ -132,7 +138,9 @@ STATUS adaptFrameNalsFromAnnexBToAvcc(PBYTE pFrameData,
     if (STATUS_SUCCEEDED(retStatus) && pAdaptedFrameDataSize != NULL) {
         // NOTE: Due to EPB removal we could in fact make the adaptation buffer smaller than the original
         // We will require at least the original size buffer.
-        *pAdaptedFrameDataSize = MAX(frameDataSize, (UINT32)(pAdaptedCurPnt - pAdaptedFrameData));
+        *pAdaptedFrameDataSize = (removeEpb && HANDLING_TRAILING_NALU_ZERO) ?
```
chehefen, line 141: shouldnt pAdaptedCurPnt - pAdaptedFrameData also account for the frameDataSize?

MushMal: In case of no EPB (which is the case where we need to split the runs for CPD for example instead of the frame NALs) the size will be smaller with the broken encoder hack. Hence, we can't use the original frame size. The adapted frame size can be smaller in this case.

MushMal: In case of no EPB (which is the case where we need to split the runs for CPD for example instead of the frame NALs) the size will be smaller with the broken encoder hack. Hence, we can't use the original frame size. The adapted frame size can be smaller in this case.

PR #69, opened 2020-06-22
Fix unupdated size for instrusmented realloc

lherman-cs:
Instrumented realloc doesn't keep track the memory size that was allocated originally.


src/utils/src/InstrumentedAllocators.c
```diff
@@ -170,6 +170,7 @@ PVOID instrumentedAllocatorsMemRealloc(PVOID ptr, SIZE_T size)
     } else {
         ATOMIC_ADD(&gInstrumentedAllocatorsTotalAllocationSize, size - existingSize);
     }
+    *pNewAlloc = size;
```
MushMal, line 173: WOOOOOOW!

PR #73, opened 2020-07-08
Add strnstr implementation for non-BSD

lherman-cs:
`STRNSTR` is defined in common def. But, unfortunately, `strnstr` is only defined in BSD. So, to make `STRNSTR` available for all platforms, this PR will add a default implementation in Rabin Karp algorithm.

`STRNSTR` is also needed for this PR, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/600.

### DIFF
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -803,7 +803,11 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 #define LTRIMSTR                   ltrimstr
 #define RTRIMSTR                   rtrimstr
 #define STRSTR                     strstr
+#ifdef strnstr
 #define STRNSTR                    strnstr
+#else
+#define STRNSTR                    defaultStrnstr
+#endif
 #define TOLOWER                    tolower
 #define TOUPPER                    toupper
 #define TOLOWERSTR                 tolowerstr
--- a/src/utils/src/String.c
+++ b/src/utils/src/String.c
@@ -450,3 +450,88 @@ CleanUp:
 
     return retStatus;
 }
+
+// This is a string search using Rabin–Karp algorithm.
+// Running time complexity:
+//   * Average: Θ(n + m)
+//   * Worst  : Θ((n−m)m)
+// Memory complexity: O(1)
+PCHAR defaultStrnstr(PCHAR haystack, PCHAR needle, SIZE_T len)
+{
+    UINT32 prime = 16777619;
+    UINT32 windowHash = 0;
+    UINT32 needleHash = 0;
+    UINT32 i;
+    UINT32 square = prime, power = 1;
+    UINT32 haystackSize, needleSize;
+
+    if (needle == NULL) {
+        return haystack;
+    }
+
+    if (haystack == NULL) {
+        return NULL;
+    }
+
+    haystackSize = (UINT32) STRNLEN(haystack, len);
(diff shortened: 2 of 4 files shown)


src/utils/tst/StringSearch.cpp
```diff
@@ -0,0 +1,60 @@
+#include "UtilTestFixture.h"
+
+class StringSearchFunctionalityTest : public UtilTestBase {
+};
+
+TEST_F(StringSearchFunctionalityTest, NegativeInvalidInput) {
```
MushMal, line 6: Add some tests for empty. Also, add cases where needle is larger than the stack

src/utils/src/String.c
```diff
@@ -449,4 +449,66 @@ STATUS tolowerupperstr(PCHAR pStr, UINT32 strLen, BOOL toUpper, PCHAR pConverted
...
+PCHAR defaultStringStrnstr(PCHAR haystack, PCHAR needle, SIZE_T len)
+{
+    UINT32 prime = 16777619;
+    UINT32 windowHash = 0;
+    UINT32 needleHash = 0;
+    UINT32 i;
+    UINT32 square = prime, power = 1;
+    UINT32 haystackSize, needleSize;
+
+    if (needle == NULL) {
```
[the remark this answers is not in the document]
lherman-cs: No need, since no loops will be executed and will go straight to the final return, `return NULL;`

src/utils/src/String.c
```diff
@@ -449,4 +449,66 @@ STATUS tolowerupperstr(PCHAR pStr, UINT32 strLen, BOOL toUpper, PCHAR pConverted
...
+        windowHash = windowHash * prime + (UINT32) haystack[i];
+        needleHash = needleHash * prime + (UINT32) needle[i];
+    }
+
+    if (windowHash == needleHash && STRNCMP(haystack, needle, needleSize) == 0) {
+        return haystack;
+    }
+
+    // find the largest power in O(log n)
+    for (i = needleSize; i > 0; i /= 2) {
```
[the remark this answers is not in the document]
lherman-cs: added a small explanation

src/utils/tst/StringSearch.cpp
lherman-cs: Makes sense! I've added the unit tests.

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -575,6 +575,21 @@ typedef struct stat STAT_STRUCT;
 
 #endif
 
+//
```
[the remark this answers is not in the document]
lherman-cs: ah makes sense.

src/utils/src/String.c
```diff
@@ -449,4 +449,89 @@ STATUS tolowerupperstr(PCHAR pStr, UINT32 strLen, BOOL toUpper, PCHAR pConverted
...
+        return haystack;
+    }
+
+    // Precompute the largest power in O(log n). 
+    //
+    // This algorithm takes an advantage for the fact that a number that's multiplied 
+    // by itself will double the power.
+    //
+    // For example:
+    //    a ^ 4 = (a ^ 2) * (a ^ 2) = a * a * a * a * a * a * a * a
```
MushMal, line 497: You have a ^ 8 in  a * a * a * a * a * a * a * a

lherman-cs: Oops, I can't count...

Issue #82, opened 2020-09-28
resetCurrentViewItemStreamStart not accounting for mkv tags


chehefen:
in https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Stream.c#L2107, it's not accounting the mkv tags. If there are tags present, it would cause invalid mkv.

Steps to repro:
- create stream
- putTag
- put a single fragment (only 1 key frame)
- reset connection
- stopsync
- freeStream

Expected result: everything works.
Actual result: handle 1 get invalid mkv.

Issue #81, opened 2020-09-28
In some scenarios, PIC is not recovering from invalid mkv error


chehefen:
When PIC encounters invalid mkv errors, it would mark view items as skipped. However, because invalid mkv error ack dont have timestamp, pic start marking from the current view item. If network is fast and the bad view item is before current, then the bad view item is never skipped. Then when PIC roll back it streams the bad view item again and never recover.

PR #84, opened 2020-10-01
Multiple improvements handling edge cases with recovery and generatin…

MushMal:
…g invalid MKV

* Adding unit test that validates that if an INVALID_MKV_DATA (or other error ACKs) are delivered on an already terminated upload handle will be ignored. This is the case when we have a new upload handle (PutMedia) being closed as there is no need to stream anything, CURL could still deliver the downstream bits. In our case, the backend will issue INVALID_MKV_DATA if the stream is closed with no data. We need to simply ignore this case.
* When there is a fragment metadata applied to the stream start, and, there is a rollback due to an error or connection being reset, the generated MKV will be corrupted as we are not handling the metadata correctly. Fixing the metadata application on a re-stream/rollback. Added unit test that pinpoints this behavior and validates.
* When we get an error ACK, in some cases like INVALID_MKV_DATA the error does not contain the timestamp that will let us map precisely to the timeline. In this case, we chose the current fragment which might not be the actual fragment the ACK is sent about. The actual "bad" fragment could be the one before. This is a situation with very fast upload. In this case, we mark the preceeding fragments for the upload handle as bad, up-to the last persisted ACK as it should be good.
* Streamlined the MKV tags generation and removed some code/logic
* Incremented the MKV generator minor version - need this as we are generating semantically the same MKV but slighly different when it comes to actual physical bytes.
* Fixed some minor misspellings and compiler warning in the unit test.

Issue #81 
Issue #82 

*Description of changes:*

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -1403,7 +1403,7 @@ STATUS getStreamData(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HANDLE uplo
             // Fill the rest of the buffer of the current view item first
             // Map the storage
             CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pKinesisVideoStream->curViewItem.viewItem.handle, (PVOID *) &pAlloc, &allocSize));
-            CHK(allocSize < MAX_UINT32 && (UINT32) allocSize <= pKinesisVideoStream->curViewItem.viewItem.length, STATUS_INVALID_ALLOCATION_SIZE);
+            CHK(allocSize < MAX_UINT32 && (UINT32) allocSize >= pKinesisVideoStream->curViewItem.viewItem.length, STATUS_INVALID_ALLOCATION_SIZE);
             size = pKinesisVideoStream->curViewItem.viewItem.length;
 
             // Validate we had allocated enough storage just in case
@@ -2102,7 +2102,7 @@ STATUS resetCurrentViewItemStreamStart(PKinesisVideoStream pKinesisVideoStream)
 
     // Get the existing frame allocation
     CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pViewItem->handle, (PVOID*) &pFrame, &allocSize));
-    CHK(allocSize < MAX_UINT32 && (UINT32) allocSize <= pViewItem->length, STATUS_INVALID_ALLOCATION_SIZE);
+    CHK(allocSize < MAX_UINT32 && (UINT32) allocSize >= pViewItem->length, STATUS_INVALID_ALLOCATION_SIZE);
 
     packagedSize = pViewItem->length;
     CHK(pFrame != NULL, STATUS_NOT_ENOUGH_MEMORY);
@@ -2159,7 +2159,7 @@ CleanUp:
             // Need to re-acquire the lock
             pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.lock);
         }
-        
+
         heapUnmap(pKinesisVideoClient->pHeap, (PVOID) pFrame);
         if (!clientLocked) {
             pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData,
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -1404,10 +1404,9 @@ STATUS getStreamData(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HANDLE uplo
             // Map the storage
             CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pKinesisVideoStream->curViewItem.viewItem.handle, (PVOID *) &pAlloc, &allocSize));
             CHK(allocSize < MAX_UINT32 && (UINT32) allocSize >= pKinesisVideoStream->curViewItem.viewItem.length, STATUS_INVALID_ALLOCATION_SIZE);
-            size = pKinesisVideoStream->curViewItem.viewItem.length;
 
             // Validate we had allocated enough storage just in case
-            CHK(pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= size, STATUS_VIEW_ITEM_SIZE_GREATER_THAN_ALLOCATION);
+            CHK(pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= allocSize, STATUS_VIEW_ITEM_SIZE_GREATER_THAN_ALLOCATION);
 
             // Copy as much as we can
             size = MIN(remainingSize, pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset);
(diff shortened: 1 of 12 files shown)


src/client/src/Stream.c
```diff
@@ -2079,38 +2096,28 @@ STATUS resetCurrentViewItemStreamStart(PKinesisVideoStream pKinesisVideoStream)
...
+    // As we are removing the MKV header, the resulting allocation size will be actually smaller
+    // We will simply copy/shift the data, including the MKV tags if any and the cluster header
+    // forward and will set the size of the allocation
 
     // Get the existing frame allocation
     CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pViewItem->handle, (PVOID*) &pFrame, &allocSize));
     CHK(allocSize < MAX_UINT32, STATUS_INVALID_ALLOCATION_SIZE);
-    packagedSize = (UINT32) allocSize;
+
+    packagedSize = pViewItem->length;
```
lherman-cs, line 2107: I think we should have a check that makes sure that `pViewItem->length <= allocSize` since this might cause an unexpected crash.

src/client/src/Stream.c
```diff
@@ -2120,11 +2127,12 @@ STATUS resetCurrentViewItemStreamStart(PKinesisVideoStream pKinesisVideoStream)
         pKinesisVideoStream->curViewItem.offset = overallSize;
     }
 
-    // Set the new length
+    // Set the new length in the view
     pViewItem->length = overallSize;
 
-    // Set the handle that will need to be freed on exit - now we should free the old one
-    allocationHandle = oldAllocationHandle;
+    // No need to set the size of the actual allocation - just unmap
```
lherman-cs, line 2133: I don't think we need to unmap here since there's another unmap at L2158.

src/client/src/Stream.c
```diff
@@ -1389,7 +1404,7 @@ STATUS getStreamData(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HANDLE uplo
             // Map the storage
             CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pKinesisVideoStream->curViewItem.viewItem.handle, (PVOID *) &pAlloc, &allocSize));
             CHK(allocSize < MAX_UINT32, STATUS_INVALID_ALLOCATION_SIZE);
-            size = (UINT32) allocSize;
+            size = pKinesisVideoStream->curViewItem.viewItem.length;
 
             // Validate we had allocated enough storage just in case
             CHK(pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= size, STATUS_VIEW_ITEM_SIZE_GREATER_THAN_ALLOCATION);
```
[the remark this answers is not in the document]
MushMal: Not sure I follow. The allocSize is a 64 bit integer whereas size is 32 so we add a check to validate that our frames are less than 32 bit in the first place.

Then, we actually use the length in the view item. Perhaps there is a diff tool issue but here is the code in this area:



CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pKinesisVideoStream->curViewItem.viewItem.handle, (PVOID *) &pAlloc, &allocSize));
            CHK(allocSize < MAX_UINT32, STATUS_INVALID_ALLOCATION_SIZE);
            size = pKinesisVideoStream->curViewItem.viewItem.length;

src/client/src/Stream.c
MushMal: I think so too indeed. The same should be with the previous case in getStreamData

src/client/src/Stream.c
MushMal: True, however, right after we unmap we set the pFrame to NULL which will make the cleanup unmapping a no-op. I generally prefer doing the cleanup if there is more stuff in the body of the function following it before the cleanup. In this case, it's nice to do that as we hold the lock. In fact, we should re-acquire the lock to do the unmapping. Will change

src/client/src/Stream.c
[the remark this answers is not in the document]
lherman-cs: Oh, I was referring to `CHK(pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= size, STATUS_VIEW_ITEM_SIZE_GREATER_THAN_ALLOCATION);`

Since size is `pKinesisVideoStream->curViewItem.viewItem.length`, the check is only validating,

pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= **pKinesisVideoStream->curViewItem.viewItem.length**

which is equal to `pKinesisVideoStream->curViewItem.offset >= 0`, which should be always true since `curViewItem.offset` is unsigned integer. (except when pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset wraps around).

src/client/src/Stream.c
```diff
@@ -1403,7 +1403,7 @@ STATUS getStreamData(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HANDLE uplo
             // Fill the rest of the buffer of the current view item first
             // Map the storage
             CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pKinesisVideoStream->curViewItem.viewItem.handle, (PVOID *) &pAlloc, &allocSize));
-            CHK(allocSize < MAX_UINT32, STATUS_INVALID_ALLOCATION_SIZE);
+            CHK(allocSize < MAX_UINT32 && (UINT32) allocSize <= pKinesisVideoStream->curViewItem.viewItem.length, STATUS_INVALID_ALLOCATION_SIZE);
```
MushMal, line 1406: D'oh, this is wrong, it should be the other way around.

src/client/src/Stream.c
```diff
@@ -2102,7 +2102,7 @@ STATUS resetCurrentViewItemStreamStart(PKinesisVideoStream pKinesisVideoStream)
 
     // Get the existing frame allocation
     CHK_STATUS(heapMap(pKinesisVideoClient->pHeap, pViewItem->handle, (PVOID*) &pFrame, &allocSize));
-    CHK(allocSize < MAX_UINT32, STATUS_INVALID_ALLOCATION_SIZE);
+    CHK(allocSize < MAX_UINT32 && (UINT32) allocSize <= pViewItem->length, STATUS_INVALID_ALLOCATION_SIZE);
```
MushMal, line 2105: same here

src/client/src/Stream.c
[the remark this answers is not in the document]
MushMal: Yes indeed, in modified code it makes lesser sense. However, this check was always an internal error catching check. In this case, it would sanitize the value of the offset. I would still leave it there though, as you say, it doesn't really add much value right now but could be a deterrent against bad code in the future that could cause weird behavior. 

We are extra careful when it comes to content store allocations. In the normal run time release heap, the allocations will be tightly packed and the overruns will not generate any crashes on the client but will corrupt the data in a weird way.

hassanctech (approved):

src/client/src/Stream.c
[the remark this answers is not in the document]
lherman-cs: Right, I was not suggesting to delete the check, but rather updating it to be 

`CHK(pKinesisVideoStream->curViewItem.viewItem.length - pKinesisVideoStream->curViewItem.offset <= allocSize, STATUS_VIEW_ITEM_SIZE_GREATER_THAN_ALLOCATION);`

This check will guarantee that we have enough memory and will not overrun.

Also, `size = pKinesisVideoStream->curViewItem.viewItem.length;` should be removed since it'll be overwritten in the line after by

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/678c208718985789e6db5f18f9437b5380290026/src/client/src/Stream.c#L1413

I probably missed something, feel free to ignore it 😊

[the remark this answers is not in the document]
MushMal: Oh, right. Uploading the change now

lherman-cs (approved):
LGTM

PR #88, opened 2020-10-20
Adding blocking wait to shutdown API for timer queue in order to clea…

MushMal:
…r the executor thread invocation. Added unit tests.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/utils/tst/TimerQueue.cpp
+++ b/src/utils/tst/TimerQueue.cpp
@@ -6,7 +6,8 @@ public:
                                     invokeCount(0),
                                     cancelAfterCount(MAX_UINT32),
                                     retStatus(STATUS_SUCCESS),
-                                    checkTimerId(TRUE)
+                                    checkTimerId(TRUE),
+                                    sleepFor(0)
     {
     }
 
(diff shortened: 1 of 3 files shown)


hassanctech (approved):
Looks good, let's wait for travis CI to succeed before merging though.

src/utils/tst/TimerQueue.cpp
```diff
@@ -733,4 +740,49 @@ TEST_F(TimerQueueFunctionalityTest, updateTimerPeriodFunctionalityTest)
...
+
+    // Wait for the next firing
+    THREAD_SLEEP(250 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+
+    // We should now block until the timer thread returns
+    curTime = GETTIME();
+    EXPECT_EQ(STATUS_SUCCESS, timerQueueShutdown(handle));
+
+    // Should have at least 3 seconds passed
+    EXPECT_LE(curTime, GETTIME());
```
lherman-cs, line 782: This will always evaluate to true even in the bad path. I think we should add an acceptable to range to pass. Probably, [`curTime`, `curTime + 3 seconds + 300 ms (for the noise)`]?

lherman-cs (commented):
Overall, it looks good to me. Only have 1 minor comment about the test.

Also, since we're changing the shutdown API to be blocking, would this break some of our higher level APIs? I'd imagine it should be working fine if we always call freeTimerQueue right after shutdown. But, I'm more concerned with the path where we call shutdown, and has another shutdown sequence from other parts before we call freeTimerQueue. The other shutdown sequences might wait the timerQueue to finish first.

src/utils/tst/TimerQueue.cpp
MushMal: Oh, of course.

MushMal: I am also thinking about the case where an API can be called after the shutdown. Need to add a check and a few tests.

Will be back with a quick change

lherman-cs (approved):
LGTM!

PR #97, opened 2020-12-10
Automatic intermittent producer handling

hassanctech:
Adding support for intermittent producer scenario.  The user does not need to worry about the scenario where frames are being sent, then there is a gap (perhaps you only want to record motion, so a period of no-motion occurs) then all of a sudden you start again, previously the backend would close the connection, now we have a timer that fires and after 20s we send a message to the back-end indicating that we're still alive so it won't close the connection.


src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1071,6 +1093,14 @@ typedef struct __ClientInfo {
...
     // Time that allowed to be elapsed between the metric loggings if enabled
     UINT64 metricLoggingPeriod;
+
+    // ------------------------------ V1 compat --------------------------
+
+    // flag for automatic handling of intermittent producer
+    AUTOMATIC_STREAMING_FLAGS automaticStreamingFlags;
+
+    // period (in hundreds of nanos) at which callback will be fired to check stream
+    UINT64 callbackPeriod;
```
MushMal, line 1103: Not sure if we need this to be public. There are certain implications if the customer sets different values - small or large. In one case, we will have odd consequences of sending EoFrs too frequently causing fragmentation whenever it's not desirable and in the other case this won't have any effect because the hard-timeout would kick in prior this. Recommend removing this and having a definition in an internal header

src/client/src/Client.c
```diff
@@ -72,6 +150,9 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
     // Set the initial stream count
     pKinesisVideoClient->streamCount = 0;
 
+    // Set the initial invocation count
+    pKinesisVideoClient->timerCallbackInvocationCount = 0;
```
MushMal, line 154: This should belong to your test

src/client/src/Client.c
```diff
@@ -146,6 +232,19 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
 
     pKinesisVideoClient->base.pStateMachine = pStateMachine;
 
+    if(pKinesisVideoClient->deviceInfo.clientInfo.automaticStreamingFlags == AUTOMATIC_STREAMING_INTERMITTENT_PRODUCER) {
+        if ( !IS_VALID_TIMER_QUEUE_HANDLE(pKinesisVideoClient->timerQueueHandle) ) {
```
MushMal, line 236: spacing is odd

src/client/src/Client.c
```diff
@@ -20,6 +20,84 @@ extern UINT32 CLIENT_STATE_MACHINE_STATE_COUNT;
...
+            if (!(pCurrStream->streamStopped || pCurrStream->streamClosed)) {
+                // Lock the Stream
+                pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData,
+                                                                 pCurrStream->base.lock);
+                streamLocked = TRUE;
+                // Check if last PutFrame is older than max timeout, if so, send EoFR, if not, do nothing
+                // Ignoring currentTime it COULD be smaller than pCurrStream->lastPutFrametimestamp
+                // Due to this method entering but waiting on stream lock from putFrame call
+                if (IS_VALID_TIMESTAMP(pCurrStream->lastPutFrameTimestamp) &&
+                    (GETTIME() - pCurrStream->lastPutFrameTimestamp) > INTERMITTENT_PRODUCER_MAX_TIMEOUT) {
```
MushMal, line 63: GETTIME can be expensive. Store the value of GETTIME before the loop and use in the iteration

src/client/src/Client.c
```diff
@@ -20,6 +20,84 @@ extern UINT32 CLIENT_STATE_MACHINE_STATE_COUNT;
...
+                    CHK_STATUS(putKinesisVideoFrame(TO_STREAM_HANDLE(pCurrStream), &eofr));
```
MushMal, line 64: Perhaps you don't want to bail out here. Maybe if (STATUS_FAILED(retStatus = putFrame())) {DLOGW("Failed to submit auto eofr with 0x%08x", retStatus);}

src/client/src/Client.c
```diff
@@ -20,6 +20,84 @@ extern UINT32 CLIENT_STATE_MACHINE_STATE_COUNT;
...
+
+                // Unlock the Stream
+                pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData,
+                                                                   pCurrStream->base.lock);
+                streamLocked = FALSE;
+            }
+        }
+    }
+
+    if(clientStreamsListLocked) {
```
MushMal, line 76: spacing is off

MushMal: Also, no need to check for clientStreamListLocked here - it always is

src/client/src/Client.c
```diff
@@ -146,6 +232,19 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
 
     pKinesisVideoClient->base.pStateMachine = pStateMachine;
 
+    if(pKinesisVideoClient->deviceInfo.clientInfo.automaticStreamingFlags == AUTOMATIC_STREAMING_INTERMITTENT_PRODUCER) {
+        if ( !IS_VALID_TIMER_QUEUE_HANDLE(pKinesisVideoClient->timerQueueHandle) ) {
+            // Create timer queue
+            CHK_STATUS(timerQueueCreate(&pKinesisVideoClient->timerQueueHandle));
+            // Store callback in client so we can override in tests
+            pKinesisVideoClient->timerCallbackFunc = checkIntermittentProducerCallback;
+            CHK_STATUS(timerQueueAddTimer(pKinesisVideoClient->timerQueueHandle, 200 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND,
```
MushMal, line 241: Define the start of the timer in an internal header and select something in a few seconds instead of millis. The application might still be doing a lot of startup/construction when this starts firing the first time which is undesirable

src/client/src/Client.c
```diff
@@ -146,6 +232,19 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
...
     pKinesisVideoClient->base.pStateMachine = pStateMachine;
 
+    if(pKinesisVideoClient->deviceInfo.clientInfo.automaticStreamingFlags == AUTOMATIC_STREAMING_INTERMITTENT_PRODUCER) {
+        if ( !IS_VALID_TIMER_QUEUE_HANDLE(pKinesisVideoClient->timerQueueHandle) ) {
+            // Create timer queue
+            CHK_STATUS(timerQueueCreate(&pKinesisVideoClient->timerQueueHandle));
+            // Store callback in client so we can override in tests
+            pKinesisVideoClient->timerCallbackFunc = checkIntermittentProducerCallback;
+            CHK_STATUS(timerQueueAddTimer(pKinesisVideoClient->timerQueueHandle, 200 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND,
+                                          pKinesisVideoClient->deviceInfo.clientInfo.callbackPeriod,
```
MushMal, line 242: Ah - I see why you are taking this in as you won't be able to reset within the test later. Here is my suggestion:

* Have a sentinel value defined as 0. 
* Add in public ClientInfo.callbackPeriod definition comments that the clients should set this field as 0. Perhaps, rename the field to reservedZero or reservedCallbackPeriod.
* In the constructor, check if it's 0 then use the defined value.

src/client/src/Client.c
```diff
@@ -205,7 +304,7 @@ STATUS createKinesisVideoClientSync(PDeviceInfo pDeviceInfo,
 
     do {
         if (pKinesisVideoClient->clientReady) {
-            DLOGV("Kinesis Video Client is Ready.");
+            DLOGW("Kinesis Video Client is Ready.");
```
MushMal, line 307: revert

src/client/src/Client.c
```diff
@@ -317,12 +421,20 @@ STATUS getKinesisVideoMetrics(CLIENT_HANDLE clientHandle, PClientMetrics pKinesi
...
+    pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.streamListLock);
+    locked = FALSE;
+
 CleanUp:
 
     if (releaseClientSemaphore) {
         semaphoreRelease(pKinesisVideoClient->base.shutdownSemaphore);
     }
 
+    if(locked) {
```
MushMal, line 434: spacing

src/client/src/Client.c
```diff
@@ -462,11 +590,27 @@ STATUS stopKinesisVideoStreamSync(STREAM_HANDLE streamHandle)
     CHK_STATUS(semaphoreAcquire(pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore, INFINITE_TIME_VALUE));
     releaseClientSemaphore = TRUE;
 
+    DLOGW("Acquired pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore");
```
MushMal, line 593: remove

src/client/src/Client.c
```diff
@@ -462,11 +590,27 @@ STATUS stopKinesisVideoStreamSync(STREAM_HANDLE streamHandle)
     CHK_STATUS(semaphoreAcquire(pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore, INFINITE_TIME_VALUE));
     releaseClientSemaphore = TRUE;
 
+    DLOGW("Acquired pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore");
+
     CHK_STATUS(semaphoreAcquire(pKinesisVideoStream->base.shutdownSemaphore, INFINITE_TIME_VALUE));
     releaseStreamSemaphore = TRUE;
 
+    DLOGW("Acquired pKinesisVideoStream->base.shutdownSemaphore");
```
MushMal, line 598: remove

src/client/src/Client.c
```diff
@@ -462,11 +590,27 @@ STATUS stopKinesisVideoStreamSync(STREAM_HANDLE streamHandle)
...
+
+
+    // lock streams list lock
+    pKinesisVideoStream->pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoStream->pKinesisVideoClient->clientCallbacks.customData,
+                                                                            pKinesisVideoStream->pKinesisVideoClient->base.streamListLock);
+    streamsListLock = TRUE;
+
     CHK_STATUS(stopStreamSync(pKinesisVideoStream));
 
+    if(streamsListLock) {
```
MushMal, line 608: spacing and remove the if statement. If you reach here then this value will be invariant will always be TRUE

src/client/src/Client.c
```diff
@@ -477,6 +621,11 @@ STATUS stopKinesisVideoStreamSync(STREAM_HANDLE streamHandle)
         semaphoreRelease(pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore);
     }
 
+    if(streamsListLock) {
```
MushMal, line 624: spacing

src/client/src/Client.c
```diff
@@ -634,6 +794,11 @@ STATUS freeKinesisVideoStream(PSTREAM_HANDLE pStreamHandle)
         semaphoreRelease(pKinesisVideoClient->base.shutdownSemaphore);
     }
 
+    if(locked) {
```
MushMal, line 797: spacing

src/client/src/Client.c
```diff
@@ -1306,13 +1479,33 @@ STATUS freeKinesisVideoClientInternal(PKinesisVideoClient pKinesisVideoClient, S
         }
     }
 
+    // unLock the streamListLock after iteration
+    if (IS_VALID_MUTEX_VALUE(pKinesisVideoClient->base.streamListLock)) {
```
MushMal, line 1483: here, you can check whether it's locked. If it is then it must have been valid too to be locked in the first place

src/client/src/Client.c
```diff
@@ -1306,13 +1479,33 @@ STATUS freeKinesisVideoClientInternal(PKinesisVideoClient pKinesisVideoClient, S
...
+
+
     // Shutdown client
     pKinesisVideoClient->clientCallbacks.clientShutdownFn(pKinesisVideoClient->clientCallbacks.customData, TO_CLIENT_HANDLE(pKinesisVideoClient));
 
     // After the shutdown
     semaphoreLock(pKinesisVideoClient->base.shutdownSemaphore);
     semaphoreWaitUntilClear(pKinesisVideoClient->base.shutdownSemaphore, CLIENT_SHUTDOWN_SEMAPHORE_TIMEOUT);
 
+    // Need to stop scheduling callbacks
```
MushMal, line 1497: Why not move this earlier and shutdown the timer queue before shutting down streams within the already held list lock?

src/client/src/Client.c
```diff
@@ -1322,6 +1515,15 @@ STATUS freeKinesisVideoClientInternal(PKinesisVideoClient pKinesisVideoClient, S
         }
     }
 
+    // unlock and free the streamListLock
+    if (IS_VALID_MUTEX_VALUE(pKinesisVideoClient->base.streamListLock)) {
```
MushMal, line 1519: You can check for locked

src/client/src/Client.c
```diff
@@ -1371,6 +1577,14 @@ STATUS freeKinesisVideoClientInternal(PKinesisVideoClient pKinesisVideoClient, S
 
 CleanUp:
 
+    // unlock and free the streamListLock
+    if (locked && IS_VALID_MUTEX_VALUE(pKinesisVideoClient->base.streamListLock)) {
```
MushMal, line 1581: If it's locked then it had been a valid mutex already

MushMal: Most importantly, you don't need to do this in the CleanUp clause. For the destructors, we don't jump to cleanup while being destructed. So, you can simply freeMutex in it's sequence like others

src/client/src/Include_i.h
```diff
@@ -261,6 +274,14 @@ typedef struct __KinesisVideoClient {
     // Total memory allocation tracker
     UINT64 totalAllocationSize;
 
+    // Timer Queue/Callback func for Automatic Intermittent Producer
+    TIMER_QUEUE_HANDLE timerQueueHandle;
+    TimerCallbackFunc timerCallbackFunc;
+    volatile UINT32 timerCallbackInvocationCount;
```
MushMal, line 280: remove

src/client/src/Stream.c
```diff
@@ -296,20 +306,32 @@ STATUS createStream(PKinesisVideoClient pKinesisVideoClient, PStreamInfo pStream
 
 CleanUp:
 
-    if (locked) {
+    if (clientLocked) {
         pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.lock);
     }
 
     if (STATUS_FAILED(retStatus) && tearDownOnError) {
+        if(!clientStreamsListLocked) {
```
MushMal, line 314: spacing

src/client/src/Stream.c
```diff
@@ -296,20 +306,32 @@ STATUS createStream(PKinesisVideoClient pKinesisVideoClient, PStreamInfo pStream
...
+            pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.streamListLock);
+            clientStreamsListLocked = TRUE;
+        }
         freeStream(pKinesisVideoStream);
     }
 
+    // Need to be under streamListLock while calling freeStream() so wait until after that check to unlock
+    if(clientStreamsListLocked) {
```
MushMal, line 322: spacing... Also, this didn't make sense. Perhaps revisit the logic of streamlist lock acquisition and release

src/client/src/Stream.c
```diff
@@ -516,28 +539,31 @@ STATUS stopStream(PKinesisVideoStream pKinesisVideoStream)
         sessionCount == 0 && // If we have active handle, then eventually one of the handle will call streamClosedFn
         !pKinesisVideoStream->metadataTracker.send &&
         !pKinesisVideoStream->eosTracker.send) {
+        DLOGW("Calling notifyStreamClosed...");
```
MushMal, line 542: remove

src/client/src/Stream.c
```diff
@@ -516,28 +539,31 @@ STATUS stopStream(PKinesisVideoStream pKinesisVideoStream)
...
         CHK_STATUS(notifyStreamClosed(pKinesisVideoStream, pUploadHandleInfo == NULL ?
                                       INVALID_UPLOAD_HANDLE_VALUE : pUploadHandleInfo->handle));
+        DLOGW("Finished Calling notifyStreamClosed...");
```
MushMal, line 545: remove after debugging is done

src/client/src/Stream.c
```diff
@@ -516,28 +539,31 @@ STATUS stopStream(PKinesisVideoStream pKinesisVideoStream)
...
+
     }
 
+    // Unlock the stream (even though it will be unlocked in the cleanup)
```
MushMal, line 549: No need for this change

src/client/src/Stream.c
```diff
@@ -594,6 +624,8 @@ STATUS stopStreamSync(PKinesisVideoStream pKinesisVideoStream)
     // release stream lock
     if (streamLocked) {
         pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+        DLOGW("Stream is unlocked");
+
```
MushMal, line 628: I don't see any need for changes in this function.

src/client/src/Stream.c
```diff
@@ -751,6 +783,9 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
 
     fixupFrame(pFrame);
 
+    // Set the last PutFrame time to current time
```
MushMal, line 786: Use this 

currentTime = pKinesisVideoClient->clientCallbacks.getCurrentTimeFn(pKinesisVideoClient->clientCallbacks.customData);

pKinesisVideoStream->lastPutFrameTimestamp = currentTIme;


The API can be expensive so we attempt to call it and store.

src/client/src/Stream.c
```diff
@@ -1930,10 +1959,6 @@ STATUS streamStartFixupOnReconnect(PKinesisVideoStream pKinesisVideoStream)
 
     pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
 
-    // Lock the stream
```
MushMal, line 1933: why changes in this function?

src/client/src/Stream.c
```diff
@@ -2079,6 +2077,10 @@ STATUS resetCurrentViewItemStreamStart(PKinesisVideoStream pKinesisVideoStream)
 
     pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
 
+    // Lock the client
```
MushMal, line 2080: This is OK

src/client/src/Stream.c
```diff
@@ -3192,6 +3191,8 @@ STATUS notifyStreamClosed(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HANDLE
     // Set the indicator of the finished stream
     pKinesisVideoStream->streamClosed = TRUE;
 
+    DLOGW("signaling stream closed condition variable");
```
MushMal, line 3194: remove once done debugging

src/client/tst/ClientTestFixture.h
```diff
@@ -781,7 +783,7 @@ class ClientTestBase : public ::testing::Test {
         return STATUS_SUCCESS;
     }
 
-    STATUS CreateStreamSync(CLIENT_HANDLE clientHandle = INVALID_STREAM_HANDLE_VALUE)
+    STATUS CreateStreamSync(CLIENT_HANDLE clientHandle = INVALID_CLIENT_HANDLE_VALUE)
```
MushMal, line 786: wow - good spotting

src/client/tst/IntermittentProducerAutomaticStreamingTest.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+};
+
+TEST_P(IntermittentProducerAutomaticStreamingTest, ValidateTimerInvokedBeforeTime) {
+    // Create new client so param value of callbackPeriod can be applied
+    ASSERT_EQ(STATUS_SUCCESS, CreateClient());
+
+    PKinesisVideoClient client = FROM_CLIENT_HANDLE(mClientHandle);
+
+    EXPECT_EQ((int)client->timerCallbackInvocationCount, 0);
```
MushMal, line 30: would recommend swapping the params as the first one is the expected result. Same for the rest of the occurences

src/client/tst/IntermittentProducerAutomaticStreamingTest.cpp
```diff
@@ -0,0 +1,192 @@
...
+    frame.frameData = temp;
+    EXPECT_EQ(STATUS_SUCCESS, putKinesisVideoFrame(mStreamHandle, &frame));
+
+    THREAD_SLEEP(200 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND + 1.25 * mDeviceInfo.clientInfo.callbackPeriod);
+
+    EXPECT_EQ((int)client->timerCallbackInvocationCount, 2);
+}
+
+/*
+This test fails, if I try to do only stopKinesisVideoStreamSync it fails there because the close
```
MushMal, line 81: We can model the true stopStreamSync in producer c SDK tests. But for here, it's harder to do. What you could do is submit the last persisted ACK or similar to trigger the stopSync to return

src/client/src/Client.c
hassanctech: It will deadlock that way, it happened during one of the existing tests.  We entered call back but another thread (in putFrame) had the stream lock and was updating `lastPutFrameTimestamp`, by the time it released the lock the saved GETTIME() value is smaller than `pCurrStream->lastPutFrameTimestamp` resulting in `(GETTIME() - pCurrStream->lastPutFrameTimestamp)` those are UINT64 but we're trying to subtract a larger value from a smaller one -- same reason why as stated in the commend I couldn't just use `currentTime` which is passed into the function.

There are a few alternatives here I can just explicitly check if the saved currTime value (or the result of GETTIME() before the loop/lock acquire) is < pCurrStream->lastPutFrameTimestamp and if it is we know there's nothing to do so we can bail. I'll make that change but just wanted to bring up the thinking here.

src/client/tst/IntermittentProducerAutomaticStreamingTest.cpp
hassanctech: will do

src/client/src/Client.c
hassanctech: Oh yes good catch! If we have multiple streams this will bail out after the first one it hits that it fails on, if we have a bad stream that's always misbehaving then subsequent ones in the list may never get eofr, will make this change!

src/client/src/Client.c
hassanctech: For consistency I kind of like the idea to just check everywhere, otherwise it's hard to spot where it's actually missing and should be there.  I suppose though if it occurs *before* the CleanUp section and the original lock is at the top of function (before any conditionals) then it will always be the case if we reached that part of execution we have the lock.  It looks like if I removed the CHK_STATUS there's no case where we can be holding onto a lock and jump to cleanup so I can in fact remove the unlock statements entirely from CleanUp, will do that.

src/client/src/Client.c
hassanctech: good point I'll make that change

src/client/src/Client.c
hassanctech: will do

src/client/src/Client.c
hassanctech: ok I remember running into issues where I had to keep checking for IS_VALID_MUTEX_VALUE during the freeKinesisVideoClientInternal call or else it would segfault, but it's possible I had other bugs then, I'll try with this simpler version.

src/client/src/Stream.c
hassanctech: freeStream actually removed it from the client streams list, hence any call to it must happen under streamlist lock.

src/client/src/Stream.c
hassanctech: fixes typo where ) was missing

I just realized I introduced a bug here by reversing the order in which the mutexes are freed (this was left over from an earlier version when i was playing with reversing the order of acquiring the mutexes) I'm reverting that

src/client/src/Stream.c
hassanctech: What is the point to use getCurrentTimeFn instead of GETTIME().  I can make this change but would like the understand the benefit of this vs just calling GETTIME()?

src/client/src/Stream.c
hassanctech: The comment says lock the stream... but there is no stream locking happening hence I removed it.

src/client/src/Client.c
hassanctech: timerQueueShutdown won't return until callbacks are all done but the callback acquires the streams list lock, so if I move it up there we'll have a deadlock because we will never be able to acquire that so I had to be careful to make sure it's *not* within that lock.

src/client/src/Client.c
MushMal: Which lock is being contended? If you refer to pClient->mTestClientMutex then this is a fast/prompt operation. Why does this deadlock?

src/client/src/Client.c
MushMal: I like the fact that you are explicitly releasing the mutex in the "happy path". My issue is with checking whether it's locked. Static code analysis tools will flag this as an invariant. Just remove the if check and simply unlock the mutex and set the locked = FALSE;

src/client/src/Client.c
MushMal: ouch.

src/client/src/Client.c
MushMal: Looks like it. This is an invariant so if it's locked then the lock must have been valid

src/client/src/Stream.c
MushMal: I saw that you are wrapping freeStream which is OK. Perhaps I commented it earlier or something like it. Ignore it

src/client/src/Stream.c
MushMal: BTW, could I ask you to resolve these convos once you fix them/resolve or any other-wise so we could keep the track

src/client/src/Stream.c
MushMal: GETTIME() maps directly to the common defs. PIC takes the platform callbacks which includes the timer. We set this in the tests, etc. Retroactively, some of these have never really found any usage in the wild and we will be removing platform callbacks from C producer unless we get more reasons to keep them. The tests could directly override the function entry points in the common defs. But, for now, let's use consistency so we can easily migrate if we do

src/client/include/com/amazonaws/kinesis/video/client/Include.h
hassanctech: Agreed to change named to `reservedCallbackPeriod` and use sentinel value, and during createKinesisVideoClient check if it was set by caller, if not then use INTERMITTENT_PRODUCER_PERIOD_DEFAULT.

src/client/src/Client.c
```diff
@@ -20,6 +20,77 @@ extern UINT32 CLIENT_STATE_MACHINE_STATE_COUNT;
...
+    STATUS retStatus = STATUS_SUCCESS;
+    PKinesisVideoClient pKinesisVideoClient = (PKinesisVideoClient) customData;
+    UINT32 i;
+    PKinesisVideoStream pCurrStream = NULL;
+    Frame eofr = EOFR_FRAME_INITIALIZER;
+
+    CHK(pKinesisVideoClient, STATUS_NULL_ARG);
+
+    // call pre hook function, useful for testing
+    if(pKinesisVideoClient->timerCallbackPreHookFunc != NULL) {
```
MushMal, line 47: Spacing is a little off in this file

src/client/src/Client.c
```diff
@@ -146,6 +226,19 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
 
     pKinesisVideoClient->base.pStateMachine = pStateMachine;
 
+    if(pKinesisVideoClient->deviceInfo.clientInfo.automaticStreamingFlags == AUTOMATIC_STREAMING_INTERMITTENT_PRODUCER) {
+        if (!IS_VALID_TIMER_QUEUE_HANDLE(pKinesisVideoClient->timerQueueHandle)) {
```
MushMal, line 232: Not sure why you check for the timer queue handle validity. This will be isolated per client object so when the client is being created, it will be unique.

src/client/src/Client.c
```diff
@@ -395,12 +506,23 @@ STATUS stopKinesisVideoStreams(CLIENT_HANDLE clientHandle)
...
+    streamsListLock = FALSE;
+
+
 CleanUp:
 
     if (releaseClientSemaphore) {
         semaphoreRelease(pKinesisVideoClient->base.shutdownSemaphore);
     }
 
+    if (streamsListLock) {
```
MushMal, line 523: Move this just above the releaseClientSemaphore. Shouldn't matter really in this case but will be future code change proof

src/client/src/Client.c
```diff
@@ -477,6 +608,11 @@ STATUS stopKinesisVideoStreamSync(STREAM_HANDLE streamHandle)
         semaphoreRelease(pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore);
     }
 
+    if(streamsListLock) {
+        pKinesisVideoStream->pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoStream->pKinesisVideoClient->clientCallbacks.customData,
```
MushMal, line 614: Do this before releasing the semaphore

src/client/src/InputValidator.c
```diff
@@ -306,27 +306,35 @@ VOID fixupClientInfo(PClientInfo pClientInfo, PClientInfo pOrigClientInfo)
...
-                pClientInfo->logMetric = pOrigClientInfo->logMetric;
-
                 break;
-
             default:
                 DLOGW("Invalid ClientInfo version");
         }
     }
+
+    if ( pClientInfo->reservedCallbackPeriod == INTERMITTENT_PRODUCER_PERIOD_SENTINEL_VALUE ) {
```
MushMal, line 335: spasing

MushMal (commented):
Minor comments

MushMal (approved):

MushMal (approved):

MushMal (approved):

PR #105, opened 2020-12-31
skip non-key frames after eofr

hassanctech:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/client/src/Client.c
+++ b/src/client/src/Client.c
@@ -68,9 +68,7 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
                 if (IS_VALID_TIMESTAMP(pCurrStream->lastPutFrameTimestamp) &&
                     currentTime > pCurrStream->lastPutFrameTimestamp
                     && (currentTime - pCurrStream->lastPutFrameTimestamp) > INTERMITTENT_PRODUCER_MAX_TIMEOUT) {
-                    if (STATUS_SUCCEEDED(retStatus = putKinesisVideoFrame(TO_STREAM_HANDLE(pCurrStream), &eofr))) {
-                        pCurrStream->lastPutFrameTimestamp = INVALID_TIMESTAMP_VALUE;
-                    } else {
+                    if (!STATUS_SUCCEEDED(retStatus = putKinesisVideoFrame(TO_STREAM_HANDLE(pCurrStream), &eofr))) {
                         DLOGW("Failed to submit auto eofr with 0x%08x, for stream name: %s", retStatus,
                               pCurrStream->streamInfo.name);
                     }
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -777,7 +777,6 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
 
     // Set the last PutFrame time to current time
     currentTime = pKinesisVideoClient->clientCallbacks.getCurrentTimeFn(pKinesisVideoClient->clientCallbacks.customData);
-    pKinesisVideoStream->lastPutFrameTimestamp = currentTime;
 
     if(pKinesisVideoStream->eofrFrame) {
         // After EOFR we again need to skip non-key frames before starting up new fragment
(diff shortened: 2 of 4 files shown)


src/client/src/Client.c
```diff
@@ -68,9 +68,7 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
                 if (IS_VALID_TIMESTAMP(pCurrStream->lastPutFrameTimestamp) &&
                     currentTime > pCurrStream->lastPutFrameTimestamp
                     && (currentTime - pCurrStream->lastPutFrameTimestamp) > INTERMITTENT_PRODUCER_MAX_TIMEOUT) {
-                    if (STATUS_SUCCEEDED(retStatus = putKinesisVideoFrame(TO_STREAM_HANDLE(pCurrStream), &eofr))) {
-                        pCurrStream->lastPutFrameTimestamp = INVALID_TIMESTAMP_VALUE;
-                    } else {
+                    if (!STATUS_SUCCEEDED(retStatus = putKinesisVideoFrame(TO_STREAM_HANDLE(pCurrStream), &eofr))) {
```
MushMal, line 71: Perhaps, here we should update the lastPutFrameTimestamp in Stream.c putFrame API when it succeeds?

src/client/src/Stream.c
```diff
@@ -777,7 +777,11 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
 
     // Set the last PutFrame time to current time
     currentTime = pKinesisVideoClient->clientCallbacks.getCurrentTimeFn(pKinesisVideoClient->clientCallbacks.customData);
-    pKinesisVideoStream->lastPutFrameTimestamp = currentTime;
+
+    if(pKinesisVideoStream->eofrFrame) {
```
MushMal, line 781: Should you move this below after the eofr is processed?

src/client/src/Stream.c
```diff
@@ -812,10 +816,11 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
     }
 
     // Check if the frames should be skipped and whether the skip flag needs to be reset
+    // If current frame is EOFR, We will package EOFR frame and we do not need to reset skip flag
     if (pKinesisVideoStream->skipNonKeyFrames) {
         if (CHECK_FRAME_FLAG_KEY_FRAME(pFrame->flags)) {
             pKinesisVideoStream->skipNonKeyFrames = FALSE;
-        } else {
+        } else if(!CHECK_FRAME_FLAG_END_OF_FRAGMENT(pFrame->flags)) {
```
MushMal, line 823: If you set the skipNonKeyFrames after you process the eofr you won't need this if clause which is harder to understand.

src/client/src/Stream.c
```diff
@@ -1112,6 +1117,11 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
         pKinesisVideoStream->diagnostics.lastFrameRateTimestamp = currentTime;
     }
 
+    if (retStatus == STATUS_SUCCESS) {
```
MushMal, line 1120: No need for the if check - if you are here then you've succeeded. This is also what I was referring to in the Client.c comment about setting the timestamp here.

MushMal (commented):
Thanks for handling this. Minor comments.

src/client/src/Client.c
hassanctech: Yep as you noted later this is exactly what I'm doing.

src/client/src/Stream.c
hassanctech: It's not the same thing right, it's possible pKinesisVideoStream->skipNonKeyFrames was already TRUE without me setting it here, in that case the else would not do the same thing.  

I think this is now easier to understand because I'm being explicit about skipping only if its not an EOFR (and also not a key frame).

src/client/src/Stream.c
hassanctech: Will change

src/client/src/Stream.c
hassanctech: Are you saying that we're relying on `pKinesisVideoStream->skipNonKeyFrames` being FALSE?
At stream creation the value is TRUE.  As soon as the first key-frame is encountered it becomes FALSE.  The only way it goes back to TRUE is if we have a bad fragment or during the executeStoppedStreamState if there's a timeout and no data available. 

Even if we could rely on it (which I'm not sure we can) I still think it's better to be clear about setting it in the beginning and being clear about skipping non key frames that come after an EOFR.

src/client/src/Stream.c
hassanctech: addressed with LN823 comment

MushMal (approved):
Looks good. You have spacing issues in the changes - not sure why this is passing the clang check. For example, missing space between if(

PR #108, opened 2021-01-04
Reset client state machine retryCount

Nomidia:
*Issue #, if available:*

*Description of changes:*


src/client/src/Stream.c
```diff
@@ -3323,6 +3323,11 @@ STATUS resetStream(PKinesisVideoStream pKinesisVideoStream)
         CHK_STATUS(resetStateMachineRetryCount(pKinesisVideoStream->base.pStateMachine));
     }
 
+    // Reset client state machine retry status
+    if (pKinesisVideoClient->base.pStateMachine != NULL) {
```
MushMal, line 3327: No need to check for the client state machine existence - is should be present

MushMal (commented):
This is an interesting case. Generally speaking, we shouldn't touch client states when dealing with the stream states as a client could have multiple streams. However, in this particular case I believe it's OK.

I have a single comment about removing the if check as it's not needed. If the client object exists then it's state machine exists too.

MushMal (approved):

Issue #106, opened 2021-01-04
[QUESTION] checkIntermittentProducerCallback(): operation returned status code: 0x52000084


Nomidia:
Continuous printing after pausing to put frames.

```
2021-01-04 08:26:02 DEBUG   postReadCallback(): Wrote 16372 bytes to Kinesis Video. Upload stream handle: 0
2021-01-04 08:26:02 DEBUG   postReadCallback(): Wrote 16372 bytes to Kinesis Video. Upload stream handle: 0
2021-01-04 08:26:02 DEBUG   postReadCallback(): Wrote 16372 bytes to Kinesis Video. Upload stream handle: 0
2021-01-04 08:26:02 DEBUG   postReadCallback(): Wrote 16372 bytes to Kinesis Video. Upload stream handle: 0
2021-01-04 08:26:02 DEBUG   postReadCallback(): Wrote 7939 bytes to Kinesis Video. Upload stream handle: 0
2021-01-04 08:26:02 DEBUG   postReadCallback(): Pausing CURL read for upload handle: 0
pKinesisVideoStream->streamInfo.streamCaps.frameOrderingMode=3, pFrameOrderCoordinator->keyFrameDetected=1
2021-01-04 08:26:20 WARN    putKinesisVideoFrame(): Failed to submit frame to Kinesis Video client. status: 0x52000084 decoding timestamp: 0 presentation timestamp: 0
2021-01-04 08:26:20 WARN    checkIntermittentProducerCallback(): Failed to submit auto eofr with 0x52000084, for stream name: UB3Ve2Pc2P3S3
2021-01-04 08:26:20 ERROR   checkIntermittentProducerCallback(): operation returned status code: 0x52000084
2021-01-04 08:26:20 ERROR   timerQueueExecutor(): operation returned status code: 0x52000084
pKinesisVideoStream->streamInfo.streamCaps.frameOrderingMode=3, pFrameOrderCoordinator->keyFrameDetected=1
2021-01-04 08:26:25 WARN    putKinesisVideoFrame(): Failed to submit frame to Kinesis Video client. status: 0x52000084 decoding timestamp: 0 presentation timestamp: 0
2021-01-04 08:26:25 WARN    checkIntermittentProducerCallback(): Failed to submit auto eofr with 0x52000084, for stream name: UB3Ve2Pc2P3S3
2021-01-04 08:26:25 ERROR   checkIntermittentProducerCallback(): operation returned status code: 0x52000084
[...]

PR #110, opened 2021-01-05
Fixing the case where getting the head and tail should return success…

MushMal:
… even if the view is empty

Issue #106 

*Description of changes:*


hassanctech (approved):
Just wanted to note this change in Stream.c is just removing dead code, in the case of end of fragment packagedMetadataSize is always 0.

PR #117, opened 2021-01-15
Correctness fix for shutdownStream in case of racing reset and free s…

MushMal:
…tream interacting at the callback provider level.

This is a necessary fix for the scenario when the continuous callback provider handling a reset on error (which is run on a separate thread) and we immediately continue and free the stream. There is a race possibility when the thread executing reset is setting the shutdown flag (as it's in the shutdown sequence) and the main free shutdown will then bail out thinking that another free has been called. 

*Issue #, if available:*

*Description of changes:*


hassanctech (approved):
For reference we've been seeing this issue reported by the Producer C Travis CI ThreadSanitizer run for a while now -- once this is merged we can update the PIC commit hash in producer C and see that race condition gone from the ThreadSanitizer report!  One step closer to being able to requiring ThreadSanitizer to succeed in the Producer C repo.

PR #123, opened 2021-03-16
Fragment aggregation

MushMal:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -350,6 +350,13 @@ STATUS freeStream(PKinesisVideoStream pKinesisVideoStream)
     // Lock the stream
     pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
 
+    // Unmap any outstanding fragment aggregator buffers
+    if (pKinesisVideoStream->fragmentAggregator.aggregateFragment &&
+        IS_VALID_ALLOCATION_HANDLE(pKinesisVideoStream->fragmentAggregator.allocationHandle) &&
+        pKinesisVideoStream->fragmentAggregator.pAllocation != NULL) {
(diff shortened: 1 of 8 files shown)


src/utils/src/FileIo.c
```diff
@@ -154,8 +153,8 @@ STATUS updateFile(PCHAR filePath, BOOL binMode, PBYTE pBuffer, UINT64 offset, UI
 
     CHK(0 == FSEEK(fp, (UINT32) offset, SEEK_SET), STATUS_INVALID_OPERATION);
 
-    for (i = 0, pCurPtr = pBuffer + offset; i < size; i++, pCurPtr++) {
-        CHK(EOF != FPUTC(*pCurPtr, fp), STATUS_WRITE_TO_FILE_FAILED);
+    for (i = 0; i < size; i++, pBuffer++) {
+        CHK(EOF != FPUTC(*pBuffer, fp), STATUS_WRITE_TO_FILE_FAILED);
```
hassanctech, line 157: This is not the same as before right?  Before we start writing at `(pBuffer + offset)`, now we start writing at `pBuffer`.

Looks like it should be a while loop anyway but if we keep the for loop let's set the range of `i` to be from `offset` to `offset + size` and then in the FPUTC we can add `i` to pBuffer.  

A number of ways to do it, but this change is different than what was there before.

src/client/src/Stream.c
```diff
@@ -795,6 +810,14 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
     if (pKinesisVideoStream->resetGeneratorOnKeyFrame && CHECK_FRAME_FLAG_KEY_FRAME(pFrame->flags)) {
         CHK_STATUS(mkvgenResetGenerator(pKinesisVideoStream->pMkvGenerator));
         pKinesisVideoStream->resetGeneratorOnKeyFrame = FALSE;
+
```
MushMal, line 813: Might want to add client locking/unlocking

PR #124, opened 2021-03-24
Fix frame rate calc

disa6302:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/562
*Description of changes:*
Previously, the frame rate reported used getTime() for calculation. The current frame rate truly is indicative of the observed frame rate, which means, it can give an idea of whether the encoder is stable or bursty, however, it is not truly indicative of the elementary frame rate which requires using PTS. This led to adding a new metric called `elementaryFrameRate` that indicates the original streams frame rate and not the rate at which the frames are produced. 

The PR also addresses very high FPS values being observed in multitrack scenario. We only calculate frameRate for video track.

API level unit test is added for Client and Stream metrics. 

TODO: 
Functional unit tests for metrics are missing. This is a bigger work item. We must simulate streaming in the tests and extract metrics to ensure values are appropriately calculated/reported.

### DIFF
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -1152,8 +1152,13 @@ struct __ClientMetrics {
     // Content view allocation size
     UINT64 totalContentViewsSize;
 
-    // Overall frame rate across the streams
+    // Overall frame rate across the streams: This is an indication of the rate at which frames are produced
+    // This is calculated using the clock value
     UINT64 totalFrameRate;
+
+    // Elementary stream frame rate in realtime/offline mode: This indicates the elementary stream frame rate
+    // This is calculated using the frame PTS
+    UINT64 totalElementaryFrameRate;
 
     // Overall transfer rate across the streams
     UINT64 totalTransferRate;
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -1156,12 +1156,14 @@ struct __ClientMetrics {
     // This is calculated using the clock value
     UINT64 totalFrameRate;
 
+    // Overall transfer rate across the streams
+    UINT64 totalTransferRate;
+
+    // V1 metrics following
+
     // Elementary stream frame rate in realtime/offline mode: This indicates the elementary stream frame rate
     // This is calculated using the frame PTS
     DOUBLE totalElementaryFrameRate;
-
-    // Overall transfer rate across the streams
-    UINT64 totalTransferRate;
 };
 
 typedef struct __ClientMetrics* PClientMetrics;
@@ -1190,9 +1192,6 @@ struct __StreamMetrics {
 
     // Last measured put frame rate in bytes per second
     DOUBLE currentFrameRate;
-
-    // Current stream's elementary frame rate.
-    DOUBLE elementaryFrameRate;
 
     // Last measured transfer rate in bytes per second
     UINT64 currentTransferRate;
@@ -1256,6 +1255,11 @@ struct __StreamMetrics {
 
     // Backend Data Plane API call latency which includes success and failure
     UINT64 dataApiCallLatency;
+
+    // V2 metrics following
+
+    // Current stream's elementary frame rate.
+    DOUBLE elementaryFrameRate;
 };
 
 typedef struct __StreamMetrics* PStreamMetrics;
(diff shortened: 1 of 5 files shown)


src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1186,6 +1191,9 @@ struct __StreamMetrics {
     // Last measured put frame rate in bytes per second
     DOUBLE currentFrameRate;
 
+    // Current stream's elementary frame rate.
+    DOUBLE elementaryFrameRate;
```
hassanctech, line 1195: This needs to go at the end with a comment
``` // V2 metrics following`

To match the existing format

File: src/client/include/com/amazonaws/kinesis/video/client/Include.h [hunk 2]
```diff
@@ -1152,9 +1152,14 @@ struct __ClientMetrics {
...
     UINT64 totalContentViewsSize;
 
-    // Overall frame rate across the streams
+    // Overall frame rate across the streams: This is an indication of the rate at which frames are produced
+    // This is calculated using the clock value
     UINT64 totalFrameRate;
 
+    // Elementary stream frame rate in realtime/offline mode: This indicates the elementary stream frame rate
+    // This is calculated using the frame PTS
+    DOUBLE totalElementaryFrameRate;
```
hassanctech, line 1161: This should go at the end with a comment about V1 metrics compat

src/client/src/Stream.c
```diff
@@ -1570,6 +1582,7 @@ STATUS getStreamMetrics(PKinesisVideoStream pKinesisVideoStream, PStreamMetrics
 
     // Store the frame rate and the transfer rate
     pStreamMetrics->currentFrameRate = pKinesisVideoStream->diagnostics.currentFrameRate;
+    pStreamMetrics->elementaryFrameRate = pKinesisVideoStream->diagnostics.elementaryFrameRate;
```
hassanctech, line 1585: You cannot do this.  This field is only available in V2 of the struct, see:  https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Stream.c#L1578

You need to add a new section at the end of it first like this to bail out early for V1 of the struct

src/client/tst/ClientApiTest.cpp
```diff
@@ -348,8 +348,51 @@ TEST_F(ClientApiTest, getKinesisVideoMetrics_Invalid)
...
+    EXPECT_NE(STATUS_SUCCESS, getKinesisVideoStreamMetrics(INVALID_STREAM_HANDLE_VALUE, NULL));
+    EXPECT_NE(STATUS_SUCCESS, getKinesisVideoStreamMetrics(INVALID_STREAM_HANDLE_VALUE, &streamMetrics));
+    // Checking unknown version
+    streamMetrics.version = STREAM_METRICS_CURRENT_VERSION + 1;
+    EXPECT_EQ(STATUS_INVALID_STREAM_METRICS_VERSION, getKinesisVideoStreamMetrics(streamHandle, &streamMetrics));
+
+    EXPECT_EQ(STATUS_SUCCESS, freeKinesisVideoStream(&streamHandle));
+}
+
+TEST_F(ClientApiTest, getStreamMetrics_Valid)
```
hassanctech, line 383: General comment about test I would test with the latest and a previous version of the struct to make sure the API doesn't fail now with an older version of the struct.

src/client/src/Client.c
```diff
@@ -385,6 +386,7 @@ STATUS getKinesisVideoMetrics(CLIENT_HANDLE clientHandle, PClientMetrics pKinesi
             CHK_STATUS(contentViewGetAllocationSize(pKinesisVideoClient->streams[i]->pView, &viewAllocationSize));
             pKinesisVideoMetrics->totalContentViewsSize += viewAllocationSize;
             pKinesisVideoMetrics->totalFrameRate += (UINT64) pKinesisVideoClient->streams[i]->diagnostics.currentFrameRate;
+            pKinesisVideoMetrics->totalElementaryFrameRate += pKinesisVideoClient->streams[i]->diagnostics.elementaryFrameRate;
```
hassanctech, line 389: Before adding this you need to check for the struct version.  Only V1 of the struct will have this field.

I noted below the tests should test correct behavior for the latest version as well as older versions of the struct which don't have this field to make sure it doesn't break since this is a public API.

disa6302: @hassanctech , 
As discussed, these are purely API tests. The testes with the tag `_Valid` test validity of all versions.

src/client/src/Stream.c
disa6302: Good catch! Fixed for both client and stream

hassanctech (approved):

Issue #129, opened 2021-08-06
deadlock on putframe


coldheartday:
I use kinesis-video-streams-producer c++ sdk to upload clips. I find  putFrameMultiTrack() will blocked and never return in some case. I checked and find a deadlock.

>Thread 1:  In offline mode, putFrameMultiTrack() will blocked for 15s when data buffer is full. In this case:
put frame->lock "**A**" [lock pFrameOrderCoordinator](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/FrameOrderCoordinator.c#L264)->[handleAvailability](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Stream.c#L853)->**blocked** and [waitCondition](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Stream.c#L1668), using lock "**B**"

>Thread 2: The timer task [checkIntermittentProducerCallback](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Client.c#L30) will call putKinesisVideoFrame() when no PutFrame() success for some time. In this case:
lock "**B**" [lock stream](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Client.c#L59) ->put frame->lock "**A**" [lock pFrameOrderCoordinator](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/FrameOrderCoordinator.c#L264) and **blocked**

Besides, I find  [checkIntermittentProducerCallback](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Client.c#L30) will [lock stream](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Client.c#L59) before call putKinesisVideoFrame(), but putframe() called by putKinesisVideoFrame() also [lock stream](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Stream.c#L757).

MushMal:
In the offline mode the putFrame API will block for some time (long timeout) until there is availability. The locks are re-entrant so it's OK for a thread to re-acquire the lock.

In this case, I believe the right approach is to disallow auto-intermittent producer in case of an offline mode. @hassanctech this is something you can look at?

[...]

PR #132, opened 2021-08-12
acquire frame order coordinator lock in this thread so guarantee same…

hassanctech:
… lock acquisition order as a media thread putKinesisVideoFrame API call

*Issue #, if available:* https://github.com/awslabs/amazon-kinesis-video-streams-pic/issues/129

*Description of changes:*

### DIFF
--- a/src/client/src/Client.c
+++ b/src/client/src/Client.c
@@ -58,7 +58,7 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
             if (NULL != pKinesisVideoClient->streams[i]) {
                 pCurrStream = pKinesisVideoClient->streams[i];
                 pFrameOrderCoordinator = pCurrStream->pFrameOrderCoordinator;
-                if (pCurrStream->streamInfo.streamCaps.frameOrderingMode == FRAME_ORDER_MODE_PASS_THROUGH) {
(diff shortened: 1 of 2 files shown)


src/client/src/Client.c
```diff
@@ -55,7 +57,14 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
         for (i = 0; i < pKinesisVideoClient->deviceInfo.streamCount; i++) {
             if (NULL != pKinesisVideoClient->streams[i]) {
                 pCurrStream = pKinesisVideoClient->streams[i];
-                // Lock the Stream
+                pFrameOrderCoordinator = pCurrStream->pFrameOrderCoordinator;
+                if (pCurrStream->streamInfo.streamCaps.frameOrderingMode != FRAME_ORDER_MODE_PASS_THROUGH) {
```
disa6302, line 61: Curious: What happens in the where it is `FRAME_ORDER_MODE_PASS_THROUGH`? Why don't we require locking for that case?

hassanctech: If you follow the code flow in the putKinesisVideoFrame API right before we acquire the frame order coordinator lock we check for the mode.  If mode is the pass through we short circuit and call putFrame immediately.  The reason for this is this mode indicates we have only 1 track.  And the whole point of frame order coordinate is to coordinate between multiple tracks.

disa6302: Ah! Ok. Makes sense now. My bad.

disa6302 (approved):

PR #136, opened 2021-08-30
Adding a generic utility for exponential backoff retries

kateyanurag:
*Issue #, if available:*
N/A

*Description of changes:*
Adding a utility for exponential backoff retries
Testing: Added new unit tests

*Travis CI*
See this PR - https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/137


src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
```diff
@@ -1453,6 +1453,140 @@ PUBLIC_API STATUS freeFileLogger();
...
+
+**************************************************************************************************/
+
+/**
+ * @brief Initializes exponential backoff state with default configuration
+ * This should be called once before calling exponentialBackoffBlockingWait.
+ *
+ * NOT THREAD SAFE.
+ */
+PUBLIC_API STATUS createExponentialBackoffStateWithDefaultConfig(PExponentialBackoffState*);
```
hassanctech, line 1557: To be consistent with the other create APIs like `timerQueueCreate` and `stackQueueCreate` and `hashTableCreate` etc. we should put `create` at the end.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
+#include "Include_i.h"
+
+//#define LOG_CLASS "XXXXX"
```
hassanctech, line 3: remove?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+
+    pExponentialBackoffState = (PExponentialBackoffState) MEMCALLOC(0x00, SIZEOF(ExponentialBackoffState));
+    CHK(pExponentialBackoffState != NULL, STATUS_NOT_ENOUGH_MEMORY);
+
+    CHK_STATUS(getDefaultExponentialBackOffConfig(&pExponentialBackoffConfig));
+
+    pExponentialBackoffState->pExponentialBackoffConfig = pExponentialBackoffConfig;
+    CHK_STATUS(resetRetryState(pExponentialBackoffState));
+
+    CleanUp:
```
hassanctech, line 59: formatting issue `CleanUp:` should be all the way left.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+
+    pExponentialBackoffConfig = (PExponentialBackoffConfig) MEMCALLOC(0x00, SIZEOF(ExponentialBackoffConfig));
+    CHK(pExponentialBackoffConfig != NULL, STATUS_NOT_ENOUGH_MEMORY);
+
+    *pExponentialBackoffConfig = *pBackoffConfig;
+
+    pExponentialBackoffState->pExponentialBackoffConfig = pExponentialBackoffConfig;
+    CHK_STATUS(resetRetryState(pExponentialBackoffState));
+
+    CleanUp:
```
hassanctech, line 93: formatting issue `CleanUp:` should be all the way left.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+            // Do nothing
+            DLOGD("Current status is BACKOFF_IN_PROGRESS");
+            break;
+        case BACKOFF_TERMINATED:
+            DLOGD("Cannot execute exponentialBackoffBlockingWait. Current status is BACKOFF_TERMINATED");
+            retStatus = STATUS_EXPONENTIAL_BACKOFF_INVALID_STATE;
+            break;
+        default:
+            DLOGD("Cannot execute exponentialBackoffBlockingWait. Unexpected state [%"PRIu64"]", pExponentialBackoffState->status);
+            retStatus = STATUS_EXPONENTIAL_BACKOFF_INVALID_STATE;
```
hassanctech, line 140: Another common pattern we use in cases like this is:

`CHK(FALSE, STATUS_EXPONENTIAL_BACKOFF_INVALID_STATE);`

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+    return (RAND() % jitterFactor) * HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+}
+
+UINT64 power(UINT32 base, UINT32 exponent) {
+    UINT64 result = 1;
+    while (exponent-- > 0) {
+        result *= base;
+    }
+    return result;
+}
```
hassanctech, line 115: What happens if base^exponent  is larger than UINT64?  Do we want to just check or limit the range of values for base/exponent to prevent this from happening if we know they actually won't span anywhere near the whole width of UINT32?  

I assume exponent will not be a big number so complexity is probably not a big concern otherwise you can divide and conquer here and reduce complexity from O(exp) to O(log(exp)).

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -460,37 +460,39 @@ typedef INT_PTR SSIZE_T, *PSSIZE_T;
...
+#define STATUS_THREAD_DEADLOCKED                        STATUS_BASE + 0x00000015
+#define STATUS_THREAD_DOES_NOT_EXIST                    STATUS_BASE + 0x00000016
+#define STATUS_JOIN_THREAD_FAILED                       STATUS_BASE + 0x00000017
+#define STATUS_WAIT_FAILED                              STATUS_BASE + 0x00000018
+#define STATUS_CANCEL_THREAD_FAILED                     STATUS_BASE + 0x00000019
+#define STATUS_THREAD_IS_NOT_JOINABLE                   STATUS_BASE + 0x0000001a
+#define STATUS_DETACH_THREAD_FAILED                     STATUS_BASE + 0x0000001b
+#define STATUS_THREAD_ATTR_INIT_FAILED                  STATUS_BASE + 0x0000001c
+#define STATUS_THREAD_ATTR_SET_STACK_SIZE_FAILED        STATUS_BASE + 0x0000001d
+#define STATUS_MEMORY_NOT_FREED                         STATUS_BASE + 0x0000001e
```
MushMal, line 493: I think these should go into the utils/Include.h

src/utils/src/ExponentialBackoffUtils.c
MushMal: I think we use this class with other components. You can define your own log class for the exponential backoff or use the utils generic one

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
+#include "Include_i.h"
+
+//#define LOG_CLASS "XXXXX"
+
+STATUS getDefaultExponentialBackOffConfig(PExponentialBackoffConfig* ppPExponentialBackoffConfig) {
```
MushMal, line 5: Consider naming the prefix setupDefault... or initializeDefault instead of get

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+    pExponentialBackoffConfig->minTimeToResetRetryState = HUNDREDS_OF_NANOS_IN_A_MILLISECOND * DEFAULT_KVS_MIN_TIME_TO_RESET_RETRY_STATE_MILLISECONDS;
+    pExponentialBackoffConfig->jitterFactor = DEFAULT_KVS_JITTER_FACTOR_MILLISECONDS;
+
+CleanUp:
+    (*ppPExponentialBackoffConfig) = pExponentialBackoffConfig;
+    LEAVES();
+    return retStatus;
+}
+
+STATUS resetRetryState(PExponentialBackoffState pExponentialBackoffState) {
```
MushMal, line 27: Consider naming of this and all other APIs to be a little more specific to avoid collisions

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    CHK(pExponentialBackoffState != NULL, STATUS_NULL_ARG);
+
+    pExponentialBackoffState->currentRetryCount = 0;
```
MushMal, line 33: MEMSETing to zero and resetting the rest might be better?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+}
+
+STATUS createExponentialBackoffState(PExponentialBackoffState* ppBackoffState, PExponentialBackoffConfig pBackoffConfig) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PExponentialBackoffState pExponentialBackoffState = NULL;
+    PExponentialBackoffConfig pExponentialBackoffConfig = NULL;
+
+    CHK(ppBackoffState != NULL, STATUS_NULL_ARG);
+    CHK(pBackoffConfig != NULL, STATUS_NULL_ARG);
```
MushMal, line 80: Combine?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+
+    pExponentialBackoffState = (PExponentialBackoffState) MEMCALLOC(0x00, SIZEOF(ExponentialBackoffState));
+    CHK(pExponentialBackoffState != NULL, STATUS_NOT_ENOUGH_MEMORY);
+
+    pExponentialBackoffConfig = (PExponentialBackoffConfig) MEMCALLOC(0x00, SIZEOF(ExponentialBackoffConfig));
```
MushMal, line 85: Consider making the config as part of the state structure. An extra allocation is an extra allocation which impacts custom allocators on embedded chips.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+        SAFE_MEMFREE(pExponentialBackoffState);
+        SAFE_MEMFREE(pExponentialBackoffConfig);
+    }
+
+    LEAVES();
+    return retStatus;
+}
+
+UINT64 getRandomJitter(UINT32 jitterFactor) {
+    return (RAND() % jitterFactor) * HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
```
MushMal, line 106: Having a low-level API, you need to ensure the params are sanitized. Here, you might have a jitterFactor being 0. I see that this is a compiler defined variable but would be nice to validate this by using CHECK_ERR(jitterFactor != 0, "Invalid jitter value"); which will crash the process with an error message

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+}
+
+UINT64 power(UINT32 base, UINT32 exponent) {
+    UINT64 result = 1;
```
MushMal, line 110: Ok, as this is a generic API, you might want to sanitize/validate the input either here or in the function that calls this.

src/utils/src/ExponentialBackoffUtils.c
MushMal: Simplest would be to enforce sane values.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+}
+
+STATUS exponentialBackoffBlockingWait(PExponentialBackoffState pRetryState) {
+    ENTERS();
+    PExponentialBackoffConfig pRetryConfig = NULL;
+    UINT64 currentTime = 0, waitTime = 0, jitter = 0;
+    STATUS retStatus = STATUS_SUCCESS;
+
+    CHK(pRetryState != NULL, STATUS_NULL_ARG);
+    CHK(pRetryState->pExponentialBackoffConfig != NULL, STATUS_NULL_ARG);
```
MushMal, line 155: Combine?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,223 @@
...
+
+    CHK_STATUS(validateAndUpdateExponentialBackoffStatus(pRetryState));
+
+    pRetryConfig = pRetryState->pExponentialBackoffConfig;
+
+    // If retries is exhausted, return error to the application
+    if (pRetryConfig->maxRetryCount != KVS_INFINITE_EXPONENTIAL_RETRIES) {
+        CHK(!(pRetryConfig->maxRetryCount == pRetryState->currentRetryCount),
+            STATUS_EXPONENTIAL_BACKOFF_RETRIES_EXHAUSTED);
+    }
```
MushMal, line 165: This might be the best to be expressed as a single line

CHK(maxRetryCount == KVS_INFINITE_EXP || maxRetry > currentRetry, STATUS_EXP_BACKOFF_EXHAUSTED);

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Apparently we've already defined LOG_CLASS in PlatformUtils.h which indirectly gets included in this file

```
#ifndef LOG_CLASS
#define LOG_CLASS "platform-utils"
#endif
```

I'll just remove that line

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: fixed

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: fixed

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Sure. I find it less intuitive though but I can change it.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: I like the idea of implementing it more efficiently using DAC. I've just moved this function to a separate MathUtils file and have added bounds on acceptable base and exponent values. Plus added a test for it.

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
kateyanurag: Ohh I see. Fixed.
Sorry I'm finding the include file structures little confusing.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Changed to `initializeDefaultExponentialBackOffConfig`

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Changed to more verbose name - `resetExponentialBackoffRetryState`
This hopefully will not collide.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Can't memset it to 0 because this struct has a config associated with it too. And it's a part of this struct.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: yep. done.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: 100% agree and thanks for pointing. Makes no sense to do one more allocation. I should not have done this in first place :(

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Sure. I've added check to sanitize the input. 
Btw CHK_ERR doesn't crash the process - 

```
#define CHK_ERR(condition, errRet, errorMessage, ...)                                                                                                \
    do {                                                                                                                                             \
        if (!(condition)) {                                                                                                                          \
            retStatus = (errRet);                                                                                                                    \
            DLOGE(errorMessage, ##__VA_ARGS__);                                                                                                      \
            goto CleanUp;                                                                                                                            \
        }                                                                                                                                            \
    } while (FALSE)
```

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: done

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: I had a hard time writing this line of code with CHK. I always get confused with the `!` in CHK macro. Although what you're suggesting is one liner, I find the if condition more intuitive to read.

kateyanurag:
@hassanctech @MushMal pushed a commit addressing feedback

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,222 @@
...
+
+UINT64 calculateWaitTime(PExponentialBackoffState pRetryState, PExponentialBackoffConfig pRetryConfig) {
+    // DEFAULT_KVS_EXPONENTIAL_FACTOR is 2 and if currentRetryCount > 64, then
+    // power(2, currentRetryCount) will overflow UINT64 buffer. In such case we
+    // always return the MAX_UINT64 value.
+    // This case will occur only if the exponential retries are configured with no
+    // upper bound on wait time. The implicit upper bound in that case will be MAX_UINT64.
+    if (pRetryState->currentRetryCount > 64) {
+        return MAX_UINT64;
+    }
```
hassanctech, line 121: I mean does it make any sense to do this?  What is being returned is a time this will be some insanely large time.  Also IF you did hit this condition and you returned MAX_UINT64, you WILL overflow down below:

`pRetryState->lastRetryWaitTime = currentTime + waitTime + jitter;`

I think we should enforce sane values here.

src/utils/src/ExponentialBackoffUtils.c
MushMal: There is a category of checks spelled as CHECK instead of CHK  - those will crash. Some implementations would only crash in DEBUG build as it evaluates to an ASSERT which is still OK

src/utils/src/ExponentialBackoffUtils.c
MushMal: It might take a little while to get used to these macros but believe me that creating those in the first place and meticulously using those allows the entire codebase to avoid a class of issues that normally plague native codebases.

In this case, you can certainly use the first if to avoid the || but try to not check the negation - that is really hard to read.

src/utils/src/ExponentialBackoffUtils.c
MushMal: +1

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,222 @@
...
+            STATUS_EXPONENTIAL_BACKOFF_RETRIES_EXHAUSTED);
+    }
+
+    // check the last retry time. If the difference between last retry and current time is greater than
+    // minTimeToResetRetryState, then reset the retry state
+    //
+    // Proceed if this is not the first retry
+    currentTime = GETTIME();
+    if (pRetryState->currentRetryCount != 0) {
+        if (currentTime - pRetryState->lastRetryWaitTime > pRetryConfig->minTimeToResetRetryState) {
```
MushMal, line 176: Combine the two if statements with a &&

src/utils/src/Math.c
```diff
@@ -0,0 +1,30 @@
+#include "Include_i.h"
+
+/**
+ * Calculates power using Divide and Conquer style - O(log exponent) complexity
+ *
+ * WARNING:
+ * The result could overflow if the base ^ power exceeds UINT64 limit.
+ * It is the callers responsibility to sanitize the values before calling this API
+ */
+UINT64 power(UINT32 base, UINT32 exponent) {
```
MushMal, line 10: My only concern on this is the name collision. Consider using a longer name perhaps.  Also, think if we want to change the API signature to return STATUS to indicate bad values passed in. For example, base being less than 2 and greater that some large value. Handling the overflow would also need to return a status code - something like invalid args or similar.

src/utils/src/Math.c
```diff
@@ -0,0 +1,30 @@
...
+            result *= base;
+        }
+        // divide exponent by 2
+        exponent = exponent >> 1;
+        // multiply base by itself
+        base = base * base;
+    }
+
+    return result;
+}
```
MushMal, line 30: \n at the end

src/utils/src/Math.c
```diff
@@ -0,0 +1,30 @@
...
+ */
+UINT64 power(UINT32 base, UINT32 exponent) {
+    UINT64 result = 1;
+
+    if (base == 0) {
+        return 0;
+    }
+
+    // loop till `exponent becomes 0
+    while (exponent) {
```
MushMal, line 18: The convention we use is to check explicitly for 0. This is done to avoid few typo errors which are hard to spot while debugging/revieweing

src/utils/src/Math.c
```diff
@@ -0,0 +1,30 @@
...
+ * It is the callers responsibility to sanitize the values before calling this API
+ */
+UINT64 power(UINT32 base, UINT32 exponent) {
+    UINT64 result = 1;
+
+    if (base == 0) {
+        return 0;
+    }
+
+    // loop till `exponent becomes 0
```
MushMal, line 17: I strongly suggest getting rid of this comment which doesn't seem to provide any value and instead add a comment describing the algorithm. I am not entirely sure why the choice of this particular algo as it's not immediately clear (simple perf might be the reason but not obvious)

src/utils/tst/MathTest.cpp
```diff
@@ -0,0 +1,12 @@
+#include "UtilTestFixture.h"
+
+class MathTest : public UtilTestBase {
+};
+
+TEST_F(MathTest, testPowerAPI)
```
MushMal, line 6: This test suite is missing overflow tests

MushMal (commented):
This is solid work

src/utils/tst/MathTest.cpp
kateyanurag: Intentionally not written those tests as we know there will be an overflow and we have not handled it. As explained above, I am biased towards keeping it simple.

src/utils/src/Math.c
kateyanurag: It's not obvious I agree. But it is faster relative to the previous implementation. I've added a comment on the top indicating it's DAC based algorithm and the complexity.

src/utils/src/Math.c
kateyanurag: okay

src/utils/src/Math.c
kateyanurag: added

src/utils/src/Math.c
kateyanurag: I can change the API name. 
I though about adding validations for input but I feel it is very ambiguous. We can have larger base smaller exponent and smaller base and a relatively larger exponent. It's pretty hard to decide what's right OR wrong. That's why I've added a warning comment and left it to the developer's discretion.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: done

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: I have added some upper limits for the config parameters. Those will ensure that we never reach insanely high wait time value (which is impractical too). We can get rid of `this check` after those validations :)

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,211 @@
+#include "Include_i.h"
+
+STATUS initializeDefaultExponentialBackOffConfig(PExponentialBackoffConfig pExponentialBackoffConfig) {
```
hassanctech, line 3: All other APIs use the casing "Backoff" instead of "BackOff" which is here, let's change this to:

`initializeDefaultExponentialBackoffConfig` so it can be consistent.

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,211 @@
...
+    CHK_STATUS(initializeDefaultExponentialBackOffConfig(&(pExponentialBackoffState->exponentialBackoffConfig)));
+    CHK_STATUS(resetExponentialBackoffRetryState(pExponentialBackoffState));
+
+CleanUp:
+    if (STATUS_SUCCEEDED(retStatus)) {
+        (*ppExponentialBackoffState) = pExponentialBackoffState;
+    } else {
+        if (pExponentialBackoffState != NULL) {
+            SAFE_MEMFREE(pExponentialBackoffState);
+        }
```
hassanctech, line 56: this if check is not needed the macro checks if(x) before attempting to free, i.e. if it's null it won't free it so we're just checking twice now.

src/utils/src/ExponentialBackoffUtils.c
hassanctech: +1 I think this change would be cleaner

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,211 @@
...
+        CHK(!(pRetryConfig->maxRetryCount == pRetryState->currentRetryCount),
+            STATUS_EXPONENTIAL_BACKOFF_RETRIES_EXHAUSTED);
+    }
+
+    // check the last retry time. If the difference between last retry and current time is greater than
+    // minTimeToResetRetryState, then reset the retry state
+    //
+    // Proceed if this is not the first retry
+    currentTime = GETTIME();
+    if (pRetryState->currentRetryCount != 0 && currentTime - pRetryState->lastRetryWaitTime > pRetryConfig->minTimeToResetRetryState) {
```
hassanctech, line 166: Is it possible that currentTime < pRetryState->lastRetryWaitTime which for unsigned values will result in a wrap around and incorrect behavior?  Should we add a check for that case?

For example there is something similar here:  https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/Client.c#L72

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: done

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: removed

src/utils/src/Math.c
hassanctech: +1 about changing the API name, I also think we need some way to enforce we will not overflow

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: Cool. Doing it.

src/utils/src/ExponentialBackoffUtils.c
kateyanurag: done

src/utils/src/Math.c
kateyanurag: Added checks to prevent overflows. And changed API name too.

kateyanurag:
@hassanctech @MushMal 
In the latest commit, I've addressed the following things - 
- Added checks to prevent overflow for power API
- Changed power API name
- Added check to ensure currentTime is always greater than lastRetryWaitTime to prevent any potential overflows
- Stylistic changes to use CHK macro wherever applicable

Please take a look.

src/utils/src/Math.c
```diff
@@ -0,0 +1,41 @@
...
+ * For ultra large base values (<=1000), the exponent limit is upto 6
+ */
+STATUS computePower(UINT64 base, UINT64 exponent, PUINT64 result) {
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT64 _result = 1;
+
+    CHK(result != NULL, STATUS_NULL_ARG);
+
+    // base and exponent values checks to prevent overflow
+    CHK((base <= 5 && exponent <= 27) ||
```
MushMal, line 18: Hmm ok.

src/utils/src/Math.c
```diff
@@ -0,0 +1,41 @@
...
+ * Calculates power using Divide and Conquer style - O(log exponent) complexity
+ *
+ * To prevent overflows -
+ * For smaller base values (<=5), the exponent limit is upto 27
+ * For larger base values (<=100), the exponent limit is upto 9
+ * For ultra large base values (<=1000), the exponent limit is upto 6
+ */
+STATUS computePower(UINT64 base, UINT64 exponent, PUINT64 result) {
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT64 _result = 1;
```
hassanctech, line 13: starting variable names with _ is a break from the established pattern can you just rename it so something else?

src/utils/src/Math.c
```diff
@@ -0,0 +1,41 @@
...
+
+    CHK(result != NULL, STATUS_NULL_ARG);
+
+    // base and exponent values checks to prevent overflow
+    CHK((base <= 5 && exponent <= 27) ||
+            (base <= 100 && exponent <= 9) ||
+            (base <= 1000 && exponent <= 6), STATUS_INVALID_ARG);
+
+    if (base == 0) {
+        return 0;
```
hassanctech, line 23: This function was modified to return `STATUS1` this needs to be updated to return the correct value I think you meant do:

`*result = 0`
`return retStatus`

?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,222 @@
...
+        currentTime > pRetryState->lastRetryWaitTime && // sanity check
+        currentTime - pRetryState->lastRetryWaitTime > pRetryConfig->minTimeToResetRetryState) {
+        CHK_STATUS(resetExponentialBackoffRetryState(pRetryState));
+        pRetryState->status = BACKOFF_IN_PROGRESS;
+    }
+
+    // Bound the exponential curve to maxWaitTime. Once we reach
+    // maxWaitTime, then we always wait for maxWaitTime time
+    // till the state is reset.
+    if (pRetryState->lastRetryWaitTime == pRetryConfig->maxWaitTime) {
```
hassanctech, line 185: Is it possible for `pRetryState->lastRetryWaitTime > pRetryConfig->maxWaitTime`
?  Seems like it would be an error, not intended, do we need to check for that is there a way that could happen?

src/utils/src/ExponentialBackoffUtils.c
```diff
@@ -0,0 +1,222 @@
...
+        CHK_STATUS(calculateWaitTime(pRetryState, pRetryConfig, &waitTime));
+        if (waitTime > pRetryConfig->maxWaitTime) {
+            waitTime = pRetryConfig->maxWaitTime;
+        }
+    }
+
+    jitter = getRandomJitter(pRetryConfig->jitterFactor);
+    THREAD_SLEEP(waitTime + jitter);
+
+    pRetryState->lastRetryWaitTime = currentTime + waitTime + jitter;
```
hassanctech, line 197: Do you need to check here that you're not setting it to a value > max wait time?

src/utils/src/Math.c
```diff
@@ -20,21 +20,22 @@ STATUS computePower(UINT64 base, UINT64 exponent, PUINT64 result) {
             (base <= 1000 && exponent <= 6), STATUS_INVALID_ARG);
 
     if (base == 0) {
-        return 0;
+        *result = 0;
```
hassanctech, line 23: Perhaps this should return `1` in the case of `0^0` (which is mathematically undefined) but all implementations I could find return either 1 or the equivalent of NaN.

hassanctech (approved):
lgtm

kateyanurag:
There's are some flaky unrelated tests which are intermittently failing in travis. I'm closing this PR in favor of a "green" travis check in [this PR](https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/137).

hassanctech:
> There's are some flaky unrelated tests which are intermittently failing in travis. I'm closing this PR in favor of a "green" travis check in [this PR](https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/137).

It looks like this is the one that you merged, not the other one?  Which I think is the right thing to do since this has the approval and I saw the CI succeeded on the merge, also the other PR is several commits behind (and is based off of the master branch instead of develop)

PR #138, opened 2021-09-02
KVS Events feature (images/notifications)

jdelapla:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -559,7 +559,8 @@ typedef enum {
     AUDIO_CODEC_ID_AAC,
     AUDIO_CODEC_ID_PCM_ALAW,
     AUDIO_CODEC_ID_PCM_MULAW,
-} AUDIO_CODEC_ID, *PAUDIO_CODEC_ID;
+} AUDIO_CODEC_ID,
+    *PAUDIO_CODEC_ID;
 
 /**
  * Video codec id
@@ -567,7 +568,8 @@ typedef enum {
 typedef enum {
     VIDEO_CODEC_ID_H264,
     VIDEO_CODEC_ID_H265,
-} VIDEO_CODEC_ID, *PVIDEO_CODEC_ID;
+} VIDEO_CODEC_ID,
+    *PVIDEO_CODEC_ID;
 
 #define GET_STREAMING_TYPE_STR(st)                                                                                                                   \
     ((st) == STREAMING_TYPE_REALTIME ? (PCHAR) "STREAMING_TYPE_REALTIME"                                                                             \
@@ -939,12 +941,12 @@ typedef enum {
 } CONTENT_STORE_PRESSURE_POLICY;
 
 typedef enum {
-    //base case
-    STREAM_EVENT_NONE = 0,
+    // base case
+    STREAM_EVENT_TYPE_NONE = 0,
     // To trigger image generation and persistence in s3
-    STREAM_EVENT_IMAGE_GENERATION,
+    STREAM_EVENT_TYPE_IMAGE_GENERATION,
     // To receive a notification that fragment has been persisted
-    STREAM_EVENT_NOTIFICATION,
+    STREAM_EVENT_TYPE_NOTIFICATION,
 } STREAM_EVENT_TYPE;
 
 /**
@@ -1437,14 +1439,14 @@ typedef struct __ServiceCallContext* PServiceCallContext;
 typedef struct __StreamEventMetadata StreamEventMetadata;
 struct __StreamEventMetadata {
     // optional s3 prefix
-    PBYTE imagePrefix;
+    PCHAR imagePrefix;
 
-    //optional optimization, stating how many pairs to be appended
+    // optional optimization, stating how many pairs to be appended
     UINT8 numberOfPairs;
 
     // optional custom data name/value pairs
-    PBYTE names[MAX_EVENT_CUSTOM_PAIRS];
-    PBYTE values[MAX_EVENT_CUSTOM_PAIRS];
+    PCHAR names[MAX_EVENT_CUSTOM_PAIRS];
+    PCHAR values[MAX_EVENT_CUSTOM_PAIRS];
 };
 
 typedef struct __StreamEventMetadata* PStreamEventMetadata;
@@ -2096,24 +2098,11 @@ PUBLIC_API STATUS getKinesisVideoStreamData(STREAM_HANDLE, UPLOAD_HANDLE, PBYTE,
 PUBLIC_API STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE, PCHAR, PCHAR, BOOL);
 
 /**
- * Inserts a "metadata" - a key/value string pair into the stream.
- *
- * NOTE: The metadata are modelled as MKV tags and are not immediately put into the stream as
- * it might break the fragment.
- * This is a limitation of MKV format as Tags are level 1 elements.
- * Instead, they will be accumulated and inserted in-between the fragments and at the end of the stream.
- *
- * MKV spec is available at: https://matroska.org/technical/specs/index.html
- *
- * Putting a "persistent" metadata will result in the metadata being inserted before every fragment.
- * The metadata can be changed by calling this function with the same name and a different value.
- * Specifying an empty string for the value for a persistent metadata will clear it and it won't
- * be applied to the consecutive fragments.
+ * Inserts a KVS event accompanied by optional metadata (key/value string pairs) into the stream.
  *
  * @param 1 STREAM_HANDLE - the stream handle.
- * @param 2 PCHAR - the metadata name.
- * @param 3 PCHAR - the metadata value.
- * @param 4 BOOL - Whether to keep applying the metadata to following fragments.
+ * @param 2 STREAM_EVENT_TYPE - the type of event
+ * @param 3 PStreamEventMetadata - pointer to struct with optional metadata
  *
  * @return Status of the function call.
  */
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -940,14 +940,11 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
-typedef enum {
-    // base case
-    STREAM_EVENT_TYPE_NONE = 0,
-    // To trigger image generation and persistence in s3
-    STREAM_EVENT_TYPE_IMAGE_GENERATION,
-    // To receive a notification that fragment has been persisted
-    STREAM_EVENT_TYPE_NOTIFICATION,
-} STREAM_EVENT_TYPE;
+// KVS Stream events
+#define STREAM_EVENT_TYPE_IMAGE_GENERATION 0x00000001L
+#define STREAM_EVENT_TYPE_NOTIFICATION     0x00000002L
+// used to iterative purposes, always keep last. NOTE have most significant bit set to 1. 0x8* not allowed
+#define STREAM_EVENT_TYPE_LAST 0x00000004L
 
 /**
  * Stream capabilities declaration
@@ -2098,15 +2095,16 @@ PUBLIC_API STATUS getKinesisVideoStreamData(STREAM_HANDLE, UPLOAD_HANDLE, PBYTE,
 PUBLIC_API STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE, PCHAR, PCHAR, BOOL);
 
 /**
- * Inserts a KVS event accompanied by optional metadata (key/value string pairs) into the stream.
+ * Inserts a KVS event(s) accompanied by optional metadata (key/value string pairs) into the stream.
+ * Multiple events can be submitted at once by using bitwise OR
  *
  * @param 1 STREAM_HANDLE - the stream handle.
- * @param 2 STREAM_EVENT_TYPE - the type of event
+ * @param 2 UINT32 - the type of event(s)
  * @param 3 PStreamEventMetadata - pointer to struct with optional metadata
  *
  * @return Status of the function call.
  */
-PUBLIC_API STATUS putKinesisVideoEventMetadata(STREAM_HANDLE, STREAM_EVENT_TYPE, PStreamEventMetadata);
+PUBLIC_API STATUS putKinesisVideoEventMetadata(STREAM_HANDLE, UINT32, PStreamEventMetadata);
 
 ////////////////////////////////////////////////////
 // Diagnostics functions
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -941,7 +941,7 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
-// KVS Stream events
+// KVS Stream events (bit flags)
 #define STREAM_EVENT_TYPE_IMAGE_GENERATION 0x00000001L
 #define STREAM_EVENT_TYPE_NOTIFICATION     0x00000002L
 // used to iterative purposes, always keep last. NOTE have most significant bit set to 1. 0x8* not allowed
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -941,11 +941,26 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
-// KVS Stream events (bit flags)
-#define STREAM_EVENT_TYPE_IMAGE_GENERATION 0x00000001L
-#define STREAM_EVENT_TYPE_NOTIFICATION     0x00000002L
-// used to iterative purposes, always keep last.
-#define STREAM_EVENT_TYPE_LAST 0x00000004L
+/**
+ * Macros checking for the stream event types
+ */
+#define CHECK_STREAM_EVENT_TYPE_IMAGE_GENERATION(f) (((f) &STREAM_EVENT_TYPE_IMAGE_GENERATION) != STREAM_EVENT_TYPE_NONE)
+#define CHECK_STREAM_EVENT_TYPE_NOTIFICATION(f)     (((f) &STREAM_EVENT_TYPE_NOTIFICATION) != STREAM_EVENT_TYPE_NONE)
+#define CHECK_STREAM_EVENT_TYPE_LAST(f)             (((f) &STREAM_EVENT_TYPE_LAST) != STREAM_EVENT_TYPE_NONE)
+
+typedef enum {
+
+    // KVS Stream events (bit flags)a
+    STREAM_EVENT_TYPE_NONE = 0,
+
+    STREAM_EVENT_TYPE_IMAGE_GENERATION = (1 << 0),
+
+    STREAM_EVENT_TYPE_NOTIFICATION = (1 << 1),
+
+    // used to iterative purposes, always keep last.
+    STREAM_EVENT_TYPE_LAST = (1 << 2),
+
+} STREAM_EVENT_TYPE;
 
 /**
  * Stream capabilities declaration
@@ -2103,7 +2118,7 @@ PUBLIC_API STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE, PCHAR, PCHAR, B
  * Multiple events can be submitted at once by using bitwise OR
  *
  * @param 1 STREAM_HANDLE - the stream handle.
- * @param 2 UINT32 - the type of event(s)
+ * @param 2 UINT32 - the type of event(s), it is suggested to use bit-wise OR combination from STREAM_EVENT_TYPE enum
  * @param 3 PStreamEventMetadata - pointer to struct with optional metadata
  *
  * @return Status of the function call.
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -267,7 +267,17 @@ extern "C" {
 /**
  * Maximal size of the metadata queue for a fragment
  */
-#define MAX_FRAGMENT_METADATA_COUNT 10
+#define MAX_FRAGMENT_METADATA_COUNT 25
+
+/**
+ * Maximum amount of the "tags" element metadata for a fragment
+ */
+#define MAX_FRAGMENT_METADATA_TAGS 10
+
+/**
+ * Max name/value pairs for custom event metadata
+ */
+#define MAX_EVENT_CUSTOM_PAIRS 5
 
 /**
  * Max length of the fragment sequence number
@@ -928,6 +938,15 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
+typedef enum {
+    //base case
+    STREAM_EVENT_NONE = 0,
+    // To trigger image generation and persistence in s3
+    STREAM_EVENT_IMAGE_GENERATION,
+    // To receive a notification that fragment has been persisted
+    STREAM_EVENT_NOTIFICATION,
+} STREAM_EVENT_TYPE;
+
 /**
  * Stream capabilities declaration
  */
@@ -1414,6 +1433,21 @@ struct __ServiceCallContext {
 };
 
 typedef struct __ServiceCallContext* PServiceCallContext;
+
+typedef struct __StreamEventMetadata StreamEventMetadata;
+struct __StreamEventMetadata {
+    // optional s3 prefix
+    PBYTE imagePrefix;
+
+    //optional optimization, stating how many pairs to be appended
+    UINT8 numberOfPairs;
+
+    // optional custom data name/value pairs
+    PBYTE names[MAX_EVENT_CUSTOM_PAIRS];
+    PBYTE values[MAX_EVENT_CUSTOM_PAIRS];
+};
+
+typedef struct __StreamEventMetadata* PStreamEventMetadata;
 
 ////////////////////////////////////////////////////
 // General callbacks definitions
@@ -2060,6 +2094,30 @@ PUBLIC_API STATUS getKinesisVideoStreamData(STREAM_HANDLE, UPLOAD_HANDLE, PBYTE,
  * @return Status of the function call.
  */
 PUBLIC_API STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE, PCHAR, PCHAR, BOOL);
+
+/**
+ * Inserts a "metadata" - a key/value string pair into the stream.
+ *
+ * NOTE: The metadata are modelled as MKV tags and are not immediately put into the stream as
+ * it might break the fragment.
+ * This is a limitation of MKV format as Tags are level 1 elements.
+ * Instead, they will be accumulated and inserted in-between the fragments and at the end of the stream.
+ *
+ * MKV spec is available at: https://matroska.org/technical/specs/index.html
+ *
+ * Putting a "persistent" metadata will result in the metadata being inserted before every fragment.
+ * The metadata can be changed by calling this function with the same name and a different value.
+ * Specifying an empty string for the value for a persistent metadata will clear it and it won't
+ * be applied to the consecutive fragments.
+ *
+ * @param 1 STREAM_HANDLE - the stream handle.
+ * @param 2 PCHAR - the metadata name.
+ * @param 3 PCHAR - the metadata value.
+ * @param 4 BOOL - Whether to keep applying the metadata to following fragments.
+ *
+ * @return Status of the function call.
+ */
+PUBLIC_API STATUS putKinesisVideoEventMetadata(STREAM_HANDLE, STREAM_EVENT_TYPE, PStreamEventMetadata);
 
 ////////////////////////////////////////////////////
 // Diagnostics functions
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -447,18 +447,19 @@ extern "C" {
 /**
  * Current versions for the public structs
  */
-#define DEVICE_INFO_CURRENT_VERSION          1
-#define CALLBACKS_CURRENT_VERSION            0
-#define STREAM_INFO_CURRENT_VERSION          2
-#define SEGMENT_INFO_CURRENT_VERSION         0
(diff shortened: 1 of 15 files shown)


src/client/src/Stream.h
```diff
@@ -78,12 +78,25 @@ extern "C" {
  */
 #define MAX_BLOCKING_PUT_WAIT 15 * HUNDREDS_OF_NANOS_IN_A_SECOND
 
+/**
+ * Maximum about of "tags" elements allowed in fragment metadata
+ */
+#define MAX_TAGS_COUNT 10
```
MushMal, line 84: Does this not conflict with other defines? Consider adding more verbosity to this

src/mkvgen/src/MkvGenerator.c
```diff
@@ -620,6 +620,168 @@ STATUS mkvgenGenerateTag(PMkvGenerator pMkvGenerator, PBYTE pBuffer, PCHAR tagNa
...
+     */
+    switch(parent) {
+        case MKV_TREE_TAGS:
+            totalOffset += MKV_TAGS_ELEMENT_SIZE_OFFSET;
+            encodedElementLength = 0x100000000000000ULL | (UINT64)(packagedSize - (totalOffset + MKV_GENERIC_ELEMENT_OFFSET));
+            PUT_UNALIGNED_BIG_ENDIAN((PINT64)(pStartPnt + totalOffset), encodedElementLength);
+            totalOffset += MKV_GENERIC_ELEMENT_OFFSET; 
+
+        case MKV_TREE_TAG:
+            totalOffset += MKV_GENERIC_ELEMENT_SIZE_OFFSET;
```
MushMal, line 745: These case statements do the same thing?

MushMal (commented):
This is solid work but without any context it's hard to review. I have added some rudimentary comments - more of a syntax level rather anything deeper/semantics level

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -928,6 +938,15 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
+typedef enum {
+    //base case
+    STREAM_EVENT_NONE = 0,
```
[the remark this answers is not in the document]
hassanctech: +1

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+    // Check if we are in the right state only if we are not in a rotation state
+    if (pKinesisVideoStream->streamState != STREAM_STATE_READY) {
+        CHK_STATUS(acceptStateMachineState(pKinesisVideoStream->base.pStateMachine,
+                                           STREAM_STATE_READY | STREAM_STATE_PUT_STREAM | STREAM_STATE_TAG_STREAM | STREAM_STATE_STREAMING |
+                                               STREAM_STATE_GET_ENDPOINT | STREAM_STATE_GET_TOKEN | STREAM_STATE_STOPPED));
+    }
+
+    // Validate if the customer is not attempting to add an internal metadata
+    if(hasMetadata) {
+        for(UINT8 iter = 0; iter < pMetadata->numberOfPairs; iter++) {
```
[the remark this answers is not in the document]
jdelapla: Iter is a for loop function, not a local function, short for iterator.
The max amount of pairs is set to 5, so I prefer to use the smallest type available, UINT8.

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+            CHK(0 != STRNCMP(AWS_INTERNAL_METADATA_PREFIX, pMetadata->names[iter], (SIZEOF(AWS_INTERNAL_METADATA_PREFIX) - 1) / SIZEOF(CHAR)), STATUS_INVALID_METADATA_NAME);
```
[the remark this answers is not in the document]
jdelapla: This is copied from the same style of function putFragmentMetdata. I lean towards keeping consistency instead of changing both.

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+        }
+        if(pMetadata->imagePrefix != NULL) {
+            CHK(0 != STRNCMP(AWS_INTERNAL_METADATA_PREFIX, pMetadata->imagePrefix, (SIZEOF(AWS_INTERNAL_METADATA_PREFIX) - 1) / SIZEOF(CHAR)), STATUS_INVALID_METADATA_NAME);
```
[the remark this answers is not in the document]
jdelapla: same

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+    neededNodes = 0;
+
+
+    // Check whether we are OK to package the metadata but do not package.
+    // create all metadata nodes
+    creatingNodes = TRUE;
+    switch(event) {
+        case STREAM_EVENT_NOTIFICATION:
+            if(firstEvent) {
+                CHK_STATUS(mkvgenGenerateTagsChain(NULL, KVSEVENT_NOTIFICATION_STRING, "", &packagedSize, MKV_TREE_TAGS));
```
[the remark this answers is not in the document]
jdelapla: I very much disagree that adding a ternary operator in function parameters improved readability over an external if/else.

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+                CHK_STATUS(mkvgenGenerateTagsChain(NULL, KVSEVENT_IMAGE_PREFIX_STRING, pMetadata->imagePrefix, &packagedSize, MKV_TREE_SIMPLE));
+                CHK_STATUS(createSerializedMetadata(KVSEVENT_IMAGE_PREFIX_STRING, pMetadata->imagePrefix, FALSE, packagedSize, event, MKV_TREE_SIMPLE, &serializedNodes[neededNodes++]));
+            }
+            break;
+        default:
+            break;
+    }
+    // Check that custom data can fit as well
+    if(hasMetadata) {
+        for(UINT8 list = 0; list < pMetadata->numberOfPairs; list++) {
```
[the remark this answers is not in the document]
jdelapla: Not a local. Just not a single character.

src/mkvgen/src/MkvGenerator.c
jdelapla: I'm stilling add comments, but these case statements are purposefully missing a break. TAG and SIMPLE have the same offsets, so their case statements are quite similar but the missing break distinguishes their functionality.

src/client/src/Client.c
```diff
@@ -862,6 +862,52 @@ STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE streamHandle, PCHAR name, P
...
+STATUS putKinesisVideoEventMetadata(STREAM_HANDLE streamHandle, STREAM_EVENT_TYPE event, PStreamEventMetadata pStreamEventMetadata)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL releaseClientSemaphore = FALSE, releaseStreamSemaphore = FALSE;
+    PKinesisVideoStream pKinesisVideoStream = FROM_STREAM_HANDLE(streamHandle);
+
+    DLOGS("Put event into an Kinesis Video stream.");
+
+    CHK(pKinesisVideoStream != NULL && pKinesisVideoStream->pKinesisVideoClient != NULL, STATUS_NULL_ARG);
+    if(pStreamEventMetadata != NULL) {
```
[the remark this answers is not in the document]
jdelapla: The event meta data is optional. It is sanitized here because this API has an expectation that the meta data will only have complete pairs, and a limited number of them.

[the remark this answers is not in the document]
jdelapla: The event meta data is optional. It is sanitized here because this API has an expectation that the meta data will only have complete pairs, and a limited number of them.

src/client/src/Stream.c
```diff
@@ -824,7 +829,8 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
 
     if (CHECK_FRAME_FLAG_END_OF_FRAGMENT(pFrame->flags)) {
         // We will append the EoFr tag and package the tags
-        CHK_STATUS(appendValidatedMetadata(pKinesisVideoStream, (PCHAR) EOFR_METADATA_NAME, (PCHAR) "", FALSE, pKinesisVideoStream->eosTracker.size));
+        CHK_STATUS(createSerializedMetadata((PCHAR) EOFR_METADATA_NAME, (PCHAR) "", FALSE, pKinesisVideoStream->eosTracker.size, STREAM_EVENT_NONE, MKV_TREE_TAGS, &pSerializedMetadata));
+        CHK_STATUS(appendValidatedMetadata(pKinesisVideoStream, pSerializedMetadata));
```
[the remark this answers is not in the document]
jdelapla: It is freed if the append fails, or it is freed after it is processed from the queue in which it was appended.

src/client/src/Client.c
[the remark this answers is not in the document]
MushMal: Most of the code assumes the semantical validation is done in the internal API. Not a biggie but it's a cleaner separation of context. Nothing prevents checking the contents in the internal API which might or might not be called from elsewhere - just a more robust approach for change-proofing in mu opinion.

src/client/src/Stream.c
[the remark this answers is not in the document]
MushMal: Nevertheless, some compiler front-ends might not handle the in-line declaration of the variable. If you see the entire core codebase has no elements of in-line variable declarations.

Also, using UINT8 (as the params are sanitized anyway) might lead to worse perf on some systems. The consistency here is that the codebase is using UINT32 for counters even for small values. WebRTC has some areas where for some reason UINT8 is being used as a counter. I think the choice of UINT8 is not a biggie so it's OK.

src/client/src/Stream.c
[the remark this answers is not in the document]
MushMal: Sigh... Would be nice to modify in both places instead of adding more of the usage.

src/client/src/Stream.c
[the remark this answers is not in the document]
MushMal: Perhaps. However, the repetition of code in these types of constructs makes the code very brittle and not future proof. You could see the entire codebase has similar constructs and it's there for a reason. 
An alternative might be to prepare the argument (with or without ternary operator) and then call the APIs without branching (aside from perf gains that non-branching provides)

src/client/src/Stream.c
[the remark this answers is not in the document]
MushMal: UIN8 list declares a local stack variable. Some compiler toolchains might break.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1415,6 +1434,21 @@ struct __ServiceCallContext {
 
 typedef struct __ServiceCallContext* PServiceCallContext;
 
+typedef struct __StreamEventMetadata StreamEventMetadata;
+struct __StreamEventMetadata {
+    // optional s3 prefix
+    PBYTE imagePrefix;
```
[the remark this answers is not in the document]
hassanctech: +1

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1415,6 +1434,21 @@ struct __ServiceCallContext {
...
+typedef struct __StreamEventMetadata StreamEventMetadata;
+struct __StreamEventMetadata {
+    // optional s3 prefix
+    PBYTE imagePrefix;
+
+    //optional optimization, stating how many pairs to be appended
+    UINT8 numberOfPairs;
+
+    // optional custom data name/value pairs
+    PBYTE names[MAX_EVENT_CUSTOM_PAIRS];
```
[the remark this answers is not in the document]
hassanctech: +1

src/client/src/Stream.c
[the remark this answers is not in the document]
hassanctech: The convention in the code today is that all local variables are declared at the top.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -928,6 +938,15 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
+typedef enum {
+    //base case
+    STREAM_EVENT_NONE = 0,
+    // To trigger image generation and persistence in s3
+    STREAM_EVENT_IMAGE_GENERATION,
+    // To receive a notification that fragment has been persisted
+    STREAM_EVENT_NOTIFICATION,
```
hassanctech, line 947: I wonder if instead of an enum if maybe these should be bit flags so that multiple options can be selected here in case in the future we will have more options here and if it's possible a customer may want to do multiple ones?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -2061,6 +2095,30 @@ PUBLIC_API STATUS getKinesisVideoStreamData(STREAM_HANDLE, UPLOAD_HANDLE, PBYTE,
...
+ * Specifying an empty string for the value for a persistent metadata will clear it and it won't
+ * be applied to the consecutive fragments.
+ *
+ * @param 1 STREAM_HANDLE - the stream handle.
+ * @param 2 PCHAR - the metadata name.
+ * @param 3 PCHAR - the metadata value.
+ * @param 4 BOOL - Whether to keep applying the metadata to following fragments.
+ *
+ * @return Status of the function call.
+ */
```
hassanctech, line 2119: The comments here are copy pasted from existing method but needs to be updated, the params are different.

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+    // Check if the stream has been stopped
+    CHK(!pKinesisVideoStream->streamStopped, STATUS_STREAM_HAS_BEEN_STOPPED);
+
+    // Lock the stream
+    pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+    streamLocked = TRUE;
+
+    if(pMetadata == NULL) {
+        hasMetadata = FALSE;
+    }
```
hassanctech, line 1874: Is there a reason this needs to be set inside the lock?  Since the value of `hasMetadata` only depends on a local variable which was passed in `pMetadata` why not just set it at the top where it's declared?

```
BOOL streamLocked = FALSE, hasMetadata = pMetadata == NULL ? FALSE : TRUE, firstEvent = TRUE, creatingNodes = FALSE;
```

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+            neededNodes++;
+            break;
+        case STREAM_EVENT_IMAGE_GENERATION:
+            neededNodes++;
+            if(hasMetadata && pMetadata->imagePrefix != NULL) {
+                neededNodes++;
+            }
+            break;
+        default:
+            break;
```
hassanctech, line 1911: Should we log if this API was called without a known event type?

src/client/src/Stream.c
```diff
@@ -1836,6 +1842,167 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+    pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+    streamLocked = FALSE;
+
+CleanUp:
+
+    //in the event we failed when creating the nodes, free the rest.
+    if(creatingNodes) {
+        for(UINT8 iter = 0; iter < neededNodes; iter++) {
+            if (serializedNodes[iter] != NULL) {
+                MEMFREE(serializedNodes[iter]);
```
hassanctech, line 1994: We should call the safe version, `SAFE_MEMFREE` here to NULL check before freeing and set to NULL when done.

src/client/src/Stream.c
```diff
@@ -3082,9 +3274,81 @@ STATUS appendValidatedMetadata(PKinesisVideoStream pKinesisVideoStream, PCHAR na
     // Set the persistent mode
     pSerializedMetadata->persistent = persistent;
 
+CleanUp:
+
+    // Free the allocation on error.
+    if (STATUS_FAILED(retStatus) && pSerializedMetadata != NULL) {
+        MEMFREE(pSerializedMetadata);
```
hassanctech, line 3281: This is OK but you can also remove the `pSerializedMetadata != NULL` check and just replace the call to MEMFREE with `SAFE_MEMFREE` which will do the null check.

src/client/src/Client.c
```diff
@@ -862,6 +862,52 @@ STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE streamHandle, PCHAR name, P
...
+    if (releaseStreamSemaphore) {
+        semaphoreRelease(pKinesisVideoStream->base.shutdownSemaphore);
+    }
+
+    if (releaseClientSemaphore) {
+        semaphoreRelease(pKinesisVideoStream->pKinesisVideoClient->base.shutdownSemaphore);
+    }
+
+    CHK_LOG_ERR(retStatus);
+    LEAVES();
```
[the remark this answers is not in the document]
jdelapla: turns out there are many functions missing this pair. Adding to all of them.

src/client/src/Stream.c
jdelapla: will do

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -928,6 +940,15 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
+typedef enum {
+    // base case
+    STREAM_EVENT_TYPE_NONE = 0,
+    // To trigger image generation and persistence in s3
+    STREAM_EVENT_TYPE_IMAGE_GENERATION,
+    // To receive a notification that fragment has been persisted
```
unicornss, line 948: can we qualify to be fragment ingestion status on persistence e.g.
STREAM_EVENT_TYPE_PERSISTED_NOTIFICATION
will help us extend later with STREAM_EVENT_TYPE_RECEIVED_NOTIFICATION for real-time getmedia path.

src/client/src/Client.c
```diff
@@ -862,13 +865,59 @@ STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE streamHandle, PCHAR name, P
...
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL releaseClientSemaphore = FALSE, releaseStreamSemaphore = FALSE;
+    PKinesisVideoStream pKinesisVideoStream = FROM_STREAM_HANDLE(streamHandle);
+
+    DLOGS("Put event into an Kinesis Video stream.");
+
+    CHK(pKinesisVideoStream != NULL && pKinesisVideoStream->pKinesisVideoClient != NULL, STATUS_NULL_ARG);
+    if (pStreamEventMetadata != NULL) {
+        CHK(pStreamEventMetadata->numberOfPairs <= MAX_EVENT_CUSTOM_PAIRS, STATUS_INVALID_ARG);
+        for (UINT8 iter = 0; iter < pStreamEventMetadata->numberOfPairs; iter++) {
```
hassanctech, line 884: Declare variable `iter` at top

src/client/src/Stream.c
```diff
@@ -1836,6 +1843,166 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+                neededNodes++;
+            }
+            break;
+        default:
+            DLOGW("Unsupported STREAM_EVENT_TYPE, %d", event);
+            break;
+    }
+    if (hasMetadata) {
+        neededNodes += pMetadata->numberOfPairs;
+    }
```
hassanctech, line 1915: It appears that the contents of this if statement can be moved up to the `if (hasMetadata` block on line 1887?  I don't see where `pMetadata->numberOfPairs` in the lines of code in between.

src/client/src/Stream.c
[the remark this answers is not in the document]
hassanctech: I'm going to have to agree with @MushMal that in this case using ternary operator is more readable.  I think when you have these code blocks repeated over and over the first thing one tends to do is to find out how they are not identical.  I think that is more apparent with the ternary operator since the only difference in this case is the MKV_TREE_* enum.  I would go further and say we should have a ternary operator for the the `KVSEVENT*` as well reducing two if/else pairs and 8 lines of code with 2 lines of code.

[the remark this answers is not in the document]
hassanctech: You have to deal with the switch so 4 lines without if/else then, I still think that's a lot better than the identical if/elses its easy to miss withs different there since out of 5/6 parameters, only 1 is different between if and else.

src/client/src/Stream.c
hassanctech: I'm thinking more about this, can we proceed if it is an unknown / unsupported event type?  Shouldn't we just bail out?

src/client/include/com/amazonaws/kinesis/video/client/Include.h
unicornss: as we discussed offline, this can be revisited for other fragment ingestion status notifications later.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1415,6 +1436,21 @@ struct __ServiceCallContext {
 
 typedef struct __ServiceCallContext* PServiceCallContext;
 
+typedef struct __StreamEventMetadata StreamEventMetadata;
+struct __StreamEventMetadata {
```
hassanctech, line 1440: We should version this struct.  For other existing versioned structs what we do is if the current default version is higher than the one the client provides we fill in defaults for the fields that don't exist in older versions.  

Right now we have an s3 prefix field which breaks encapsulation but also is specific to an event type, future event types may also have other things specific to them that need to be specified in this struct.

src/client/src/Stream.c
```diff
@@ -1836,6 +1843,164 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+        }
+    } else {
+        ++pKinesisVideoStream->metadataTracker.eventIndex;
+
+        for (iter = 0; iter < neededNodes; iter++) {
+            CHK_STATUS(insertValidatedMetadata(pKinesisVideoStream, serializedNodes[iter], pKinesisVideoStream->metadataTracker.eventIndex + iter));
+        }
+    }
+
+    // Unlock the stream (even though it will be unlocked in the cleanup
```
hassanctech, line 1983: closed parenthesis in the comment

src/client/src/Stream.c
```diff
@@ -2745,7 +2912,8 @@ STATUS checkStreamingTokenExpiration(PKinesisVideoStream pKinesisVideoStream)
     // In case of failure with get endpoint (i.e. if client state machine fails at auth,
```
hassanctech, line 2912: need to close this paren in the comments

src/client/include/com/amazonaws/kinesis/video/client/Include.h
jdelapla: added versioning in last commit

src/client/src/Stream.c
jdelapla: added, though it's 3 lines below.

src/client/src/Stream.c
jdelapla: done

src/client/src/Stream.c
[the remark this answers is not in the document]
jdelapla: With the most recent change from enum-> bit shifted defines it became necessary to add another flag to be checked & changed, thus a ternary operator in the function is not longer sufficient.

src/client/src/Stream.c
```diff
@@ -1836,6 +1843,153 @@ STATUS setNalAdaptationFlags(PKinesisVideoStream pKinesisVideoStream, UINT32 nal
...
+        }
+        if (pMetadata->imagePrefix != NULL) {
+            CHK(0 != STRNCMP(AWS_INTERNAL_METADATA_PREFIX, pMetadata->imagePrefix, (SIZEOF(AWS_INTERNAL_METADATA_PREFIX) - 1) / SIZEOF(CHAR)),
+                STATUS_INVALID_METADATA_NAME);
+        }
+        neededNodes += pMetadata->numberOfPairs;
+    }
+
+    // Ensure we don't have more than MAX size of the metadata queue
+    if (event & STREAM_EVENT_TYPE_NOTIFICATION) {
```
hassanctech, line 1900: Can you create macros for this like we have for frame flags, see:

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/mkvgen/include/com/amazonaws/kinesis/video/mkvgen/Include.h#L230-L233

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -928,6 +941,12 @@ typedef enum {
     CONTENT_STORE_PRESSURE_POLICY_DROP_TAIL_ITEM = 1,
 } CONTENT_STORE_PRESSURE_POLICY;
 
+// KVS Stream events (bit flags)
+#define STREAM_EVENT_TYPE_IMAGE_GENERATION 0x00000001L
+#define STREAM_EVENT_TYPE_NOTIFICATION     0x00000002L
+// used to iterative purposes, always keep last.
+#define STREAM_EVENT_TYPE_LAST 0x00000004L
+
```
hassanctech, line 949: Sorry for the confusion but instead of making this complete untyped I think we should have it still have an event type just with those values, same as how we do for the frame flags, see for example:  

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/mkvgen/include/com/amazonaws/kinesis/video/mkvgen/Include.h#L176-L201

src/mkvgen/src/MkvGenerator.c
```diff
@@ -620,6 +620,168 @@ STATUS mkvgenGenerateTag(PMkvGenerator pMkvGenerator, PBYTE pBuffer, PCHAR tagNa
...
+
+        case MKV_TREE_TAG:
+            packagedSize += MKV_GENERIC_ELEMENT_SIZE_OFFSET + MKV_GENERIC_ELEMENT_OFFSET;
+
+        case MKV_TREE_SIMPLE:
+            packagedSize += MKV_GENERIC_ELEMENT_SIZE_OFFSET + MKV_GENERIC_ELEMENT_OFFSET;
+            break;
+
+        default:
+            break;
```
hassanctech, line 657: Is this an error? Do we need to log or bail out?

src/mkvgen/src/MkvGenerator.c
```diff
@@ -620,6 +620,168 @@ STATUS mkvgenGenerateTag(PMkvGenerator pMkvGenerator, PBYTE pBuffer, PCHAR tagNa
...
+        case MKV_TREE_SIMPLE:
+            encodedLen = MKV_TAG_SIMPLE_BITS_SIZE;
+            CHK(bufferSize >= encodedLen, STATUS_NOT_ENOUGH_MEMORY);
+            MEMCPY(pCurrentPnt, MKV_TAG_SIMPLE_BITS, encodedLen);
+            bufferSize -= encodedLen;
+            pCurrentPnt += encodedLen;
+            break;
+
+        default:
+            break;
```
hassanctech, line 697: Is this an error? Do we need to log or bail out?

src/mkvgen/src/MkvGenerator.c
```diff
@@ -620,6 +620,168 @@ STATUS mkvgenGenerateTag(PMkvGenerator pMkvGenerator, PBYTE pBuffer, PCHAR tagNa
...
+
+        case MKV_TREE_SIMPLE:
+            totalOffset += MKV_GENERIC_ELEMENT_SIZE_OFFSET;
+            encodedElementLength = 0x100000000000000ULL | (UINT64)(packagedSize - (totalOffset + MKV_GENERIC_ELEMENT_OFFSET));
+            PUT_UNALIGNED_BIG_ENDIAN((PINT64)(pStartPnt + totalOffset), encodedElementLength);
+            totalOffset += MKV_GENERIC_ELEMENT_OFFSET;
+
+            break;
+        default:
+            break;
```
hassanctech, line 759: Is this an error? Do we need to log or bail out?

hassanctech:
This test `StreamTokenRotationTest.rotationWithAwaitingCheck` is failing across all platforms in the CI, please check, all other tests pass.

src/client/src/Client.c
```diff
@@ -862,13 +865,66 @@ STATUS putKinesisVideoFragmentMetadata(STREAM_HANDLE streamHandle, PCHAR name, P
...
+        for (iter = 0; iter < pStreamEventMetadata->numberOfPairs; iter++) {
+            CHK(pStreamEventMetadata->names[iter] != NULL && pStreamEventMetadata->values[iter] != NULL, STATUS_NULL_ARG);
+        }
+        pStreamEventMetadata->version = STREAM_EVENT_METADATA_CURRENT_VERSION;
+    }
+
+    // check for valid events
+    if (event >= STREAM_EVENT_TYPE_LAST || event == STREAM_EVENT_TYPE_NONE) {
+        CHK(FALSE, STATUS_INVALID_ARG);
+    }
```
hassanctech, line 894: This check could be before the `if (pStreamEventMetadata != NULL)` to bail out early right after the null checks.

src/mkvgen/src/MkvGenerator.c
```diff
@@ -620,6 +620,174 @@ STATUS mkvgenGenerateTag(PMkvGenerator pMkvGenerator, PBYTE pBuffer, PCHAR tagNa
...
+    bufferSize = *pSize;
+
+    // Check the size and copy the structure first
+    switch (parent) {
+        case MKV_TREE_TAGS:
+            encodedLen = MKV_TAGS_BITS_SIZE;
+            CHK(bufferSize >= encodedLen, STATUS_NOT_ENOUGH_MEMORY);
+            MEMCPY(pCurrentPnt, MKV_TAGS_BITS, encodedLen);
+            bufferSize -= encodedLen;
+            pCurrentPnt += encodedLen;
```
hassanctech, line 680: You can move out of the switch since it's the same 2 lines in all cases.

hassanctech (approved):
lgtm

nice work!!!

only higher level comment is to add another unit test around the case which was fixed in the last commit.

src/client/src/Client.c
jdelapla: It could be, but those lines are also checking the validity of the parameters, so the ordering is pretty insignificant.

PR #143, opened 2021-09-29
Adding error handler in PIC state machine and updating client/stream state machines to use the error handler on state transitions

kateyanurag:
*Issue #, if available:*

*Description of changes:*

The commit includes following -
(1) Refactoring Exponential backoff utility by changing name(s) to use 'RetryStrategy'
(2) Minor logging improvements in  Exponential backoff utility
(3) Adding KVS retry strategy and related callbacks to be used in client and streams
(4) Adding a new error handler in PIC state machine
(5) Updating KVS client and stream state machines to use PIC error handler and exponential backoff retry strategy on encountering error while state transitions
(6) Unit tests for state machine changes


src/state/include/com/amazonaws/kinesis/video/state/Include.h
```diff
@@ -64,6 +64,16 @@ typedef STATUS (*GetNextStateFunc)(UINT64, PUINT64);
...
 
+/**
+ * State's error handler function. Called before state machine
+ * transitions to a different state.
+ *
+ * @param 1 - IN - Custom data passed in
+ *
+ * @return Status of the callback
+ */
+typedef STATUS (*ErrorHandlerStateFunc)(UINT64);
```
MushMal, line 75: This seem to be not an error handler. It's more of the state transition function. We have a very similar concept in the Signaling client in WebRTC.

Recommend changing the name of StateMachineTransitionFunc or something similar.

src/state/src/State.c
```diff
@@ -158,6 +158,12 @@ STATUS stepStateMachine(PStateMachine pStateMachine)
         pStateMachineImpl->context.retryCount = 0;
         pStateMachineImpl->context.time = time;
 
+        // Call the state machine error handler
+        // This call could be a no-op if the state transition is happening for SERVICE_CALL_RESULT_OK code
+        if (pStateMachineImpl->context.pCurrentState->errorHandlerStateFunc != NULL) {
+            pStateMachineImpl->context.pCurrentState->errorHandlerStateFunc(pStateMachineImpl->customData);
```
MushMal, line 164: The state transition function returns a status. You might want to return the status from here if it's not STATUS_SUCCESS - have a way to control the state machine execution from the "hook"

src/client/src/Client.c
```diff
@@ -94,6 +94,57 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
     return retStatus;
 }
 
+STATUS configureClientWithRetryStrategy(PKinesisVideoClient pKinesisVideoClient, KVS_RETRY_STRATEGY_TYPE kvsRetryStrategyType) {
```
MushMal, line 97: The declaration of this fn needs to go into the internal header file.

Most importantly, design of the API... You can take a moniker of the action (an enum defining what to do) or the action itself. The first is more compact and easy to use for a client interface but is very limited and hard to expand. The latter, providing the actual functions, is more flexible. Taking that this is an internal function, you could go with both approaches. I prefer the latter for testing purposes where the test could control the setting of the functions in mid-flight for example.

src/client/src/Client.c
```diff
@@ -94,6 +94,57 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
...
+        pKinesisVideoClient->kVSRetryStrategy.pRetryStrategy = pRetryStrategy;
+    }
+
+CleanUp:
+
+    LEAVES();
+    return retStatus;
+}
+
+STATUS freeClientRetryStrategy(PKinesisVideoClient pKinesisVideoClient) {
```
MushMal, line 129: Need a declaration of the API in the internal header. Most importantly, the name prefix "free" implies deallocation of resources. Perhaps deinitClientRetryStrategy?

src/client/src/Client.c
```diff
@@ -94,6 +94,57 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
...
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    CHK(pKinesisVideoClient != NULL, STATUS_NULL_ARG);
+    CHK(pKinesisVideoClient->kVSRetryStrategy.pRetryStrategy != NULL, STATUS_SUCCESS);
+
+    if (pKinesisVideoClient->kVSRetryStrategy.freeRetryStrategyFn != NULL) {
```
MushMal, line 136: Combine both in a single line

CHK(pKinesisVideoClient->kvsRetryStrategy.pRetryStrategy != NULL && pKinesisVideoClient->kvsRetryStrategy.FreeRetryStrategyFn != NULL, retStatus);

src/client/src/ClientState.c
```diff
@@ -9,24 +9,95 @@
...
-     SERVICE_CALL_MAX_RETRY_COUNT, STATUS_GET_CLIENT_TOKEN_CALL_FAILED},
-    {CLIENT_STATE_PROVISION, CLIENT_STATE_AUTH | CLIENT_STATE_PROVISION, fromProvisionClientState, executeProvisionClientState,
-     SERVICE_CALL_MAX_RETRY_COUNT, STATUS_CLIENT_PROVISION_CALL_FAILED},
-    {CLIENT_STATE_CREATE, CLIENT_STATE_PROVISION | CLIENT_STATE_GET_TOKEN | CLIENT_STATE_AUTH | CLIENT_STATE_CREATE, fromCreateClientState,
-     executeCreateClientState, SERVICE_CALL_MAX_RETRY_COUNT, STATUS_CREATE_CLIENT_CALL_FAILED},
-    {CLIENT_STATE_TAG_CLIENT, CLIENT_STATE_CREATE | CLIENT_STATE_TAG_CLIENT | CLIENT_STATE_READY, fromTagClientState, executeTagClientState,
-     SERVICE_CALL_MAX_RETRY_COUNT, STATUS_TAG_CLIENT_CALL_FAILED},
-    {CLIENT_STATE_READY, CLIENT_STATE_GET_TOKEN | CLIENT_STATE_AUTH | CLIENT_STATE_TAG_CLIENT | CLIENT_STATE_CREATE | CLIENT_STATE_READY,
-     fromReadyClientState, executeReadyClientState, INFINITE_RETRY_COUNT_SENTINEL, STATUS_CLIENT_READY_CALLBACK_FAILED},
+    {
```
MushMal, line 12: Assuming the content of the state machine states hasn't changed - hard to tell due to formatting.

src/client/src/ClientState.c
```diff
@@ -9,24 +9,95 @@
...
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    PKinesisVideoClient pKinesisVideoClient = CLIENT_FROM_CUSTOM_DATA(customData);
+    PKVSRetryStrategy pKVSRetryStrategy = &(pKinesisVideoClient->kVSRetryStrategy);
+
+    CHK(pKinesisVideoClient->base.result != SERVICE_CALL_RESULT_OK, STATUS_SUCCESS);
+    CHK(pKVSRetryStrategy != NULL &&
+        pKVSRetryStrategy->pRetryStrategy != NULL &&
+        pKVSRetryStrategy->executeRetryStrategyFn != NULL, STATUS_SUCCESS);
```
MushMal, line 89: Combine all in a single CHK with && and return retStatus (recommend)

src/client/src/Include_i.h
```diff
@@ -280,6 +280,9 @@ typedef struct __KinesisVideoClient {
     // ID for timer created to wake and check if streams have incoming data
     UINT32 timerId;
 
+    // Retry strategy for the client and all the streams under it
+    KVSRetryStrategy kVSRetryStrategy;
```
MushMal, line 284: Odd naming. recommend simply retryStrategy

src/client/src/StreamState.c
```diff
@@ -9,36 +9,129 @@
...
+    STATUS retStatus = STATUS_SUCCESS;
+
+    PKinesisVideoStream pKinesisVideoStream = STREAM_FROM_CUSTOM_DATA(customData);
+    PKinesisVideoClient pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
+    PKVSRetryStrategy pKVSRetryStrategy = &(pKinesisVideoClient->kVSRetryStrategy);
+
+    CHK(pKinesisVideoStream->base.result != SERVICE_CALL_RESULT_OK, STATUS_SUCCESS);
+    CHK(pKVSRetryStrategy != NULL &&
+        pKVSRetryStrategy->pRetryStrategy != NULL &&
+        pKVSRetryStrategy->executeRetryStrategyFn != NULL, STATUS_SUCCESS);
```
MushMal, line 123: Similar comment as in client state

src/client/src/ClientState.c
```diff
@@ -9,24 +9,95 @@
...
+    }
 };
 
 UINT32 CLIENT_STATE_MACHINE_STATE_COUNT = SIZEOF(CLIENT_STATE_MACHINE_STATES) / SIZEOF(StateMachineState);
 
+STATUS defaultClientStateMachineErrorHandler(UINT64 customData /* customData should be PKinesisVideoClient */) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    PKinesisVideoClient pKinesisVideoClient = CLIENT_FROM_CUSTOM_DATA(customData);
```
MushMal, line 83: Might want to check if pKinesisVideoClient != NULL

src/client/src/StreamState.c
```diff
@@ -9,36 +9,129 @@
...
 ///////////////////////////////////////////////////////////////////////////
 // State machine callback functions
 ///////////////////////////////////////////////////////////////////////////
+
+
+STATUS defaultStreamStateMachineErrorHandler(UINT64 customData /* customData should be PKinesisVideoStream */) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    PKinesisVideoStream pKinesisVideoStream = STREAM_FROM_CUSTOM_DATA(customData);
```
MushMal, line 116: Check for stream object not being NULL and the client not being NULL

src/state/include/com/amazonaws/kinesis/video/state/Include.h
MushMal: Another question is whether this belongs to the public header. So far, it seems to be used in the internal functionality. Is the idea to expose this API to C producer/WebRTC?

src/state/include/com/amazonaws/kinesis/video/state/Include.h
```diff
@@ -73,6 +83,7 @@ struct __StateMachineState {
     UINT64 acceptStates;
     GetNextStateFunc getNextStateFn;
     ExecuteStateFunc executeStateFn;
+    ErrorHandlerStateFunc errorHandlerStateFunc;
```
MushMal, line 86: This API was internal API until we needed to expose it to common library and webrtc. Hence, we didn't version the structure. Need to be careful with the deployment of the changes.

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -0,0 +1,251 @@
...
+        (*ppRetryStrategy) = (PRetryStrategy)pExponentialBackoffRetryStrategyState;
+    } else {
+        SAFE_MEMFREE(pExponentialBackoffRetryStrategyState);
+    }
+
+    LEAVES();
+    return retStatus;
+}
+
+STATUS inRange(UINT64 value, UINT64 low, UINT64 high) {
```
MushMal, line 62: I am afraid the naming of this internal function might conflict with other libraries.

MushMal: Moreover, I would recommend implementing this as a macro (with the name in capitals)

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -0,0 +1,251 @@
...
+            pExponentialBackoffRetryStrategyState->status = BACKOFF_IN_PROGRESS;
+            DLOGD("Status changed from BACKOFF_NOT_STARTED to BACKOFF_IN_PROGRESS");
+            break;
+        case BACKOFF_IN_PROGRESS:
+            // Do nothing
+            DLOGD("Current status is BACKOFF_IN_PROGRESS");
+            break;
+        case BACKOFF_TERMINATED:
+            DLOGD("Cannot execute exponentialBackoffBlockingWait. Current status is BACKOFF_TERMINATED");
+            CHK(FALSE, STATUS_EXPONENTIAL_BACKOFF_INVALID_STATE);
```
MushMal, line 156: You could use CHK_ERR in these cases to output the text in the error level if you want

src/state/include/com/amazonaws/kinesis/video/state/Include.h
kateyanurag: I agree that this is technically a state transition function, but we intend to use this for executing some logic for state transitions on errors. StateMachineTransitionFunc sounds more like we need to perform "some" logic while transitioning states. To me adding anything at this layer while transitioning state is a bit risky and we just intend to do error check and add wait time to avoid high level tight loops. So I suggest to stick with ErrorHandlerStateFunc name for it to be very specific.
Thoughts? 

We do not intend to expose this in public APIs. Exposing this to the customers might not be super useful. Customers should be abstracted from state machine implementation. Also exposing such low level APIs might lead customers doesn't seem right (unless there's a strong reason). This is not a one way door though. We can always expose these APIs later if need arise.

src/state/src/State.c
kateyanurag: Yeah sure. Wrapped it in CHK_STATUS macro.

src/client/src/Client.c
kateyanurag: This is actually a private function. In the newer commit, I've added a client callback to define retry strategy. With the existing approach, it is nearly impossible to unit test this code. So I had to add a callback in client. Please check the newer code. That might answer your concern.

src/client/src/Client.c
kateyanurag: Sure I've added it in the internal header

src/client/src/Client.c
kateyanurag: done

src/client/src/ClientState.c
kateyanurag: I had a hard look at this before sending PR. I can promise nothing changed. I understand that this is an extra change but it was pretty hard to read the state machine. So just formatted it to make it look cleaner.

src/client/src/ClientState.c
kateyanurag: done

src/client/src/Include_i.h
kateyanurag: Intentionally added KVS as a prefix to avoid any potential conflicts. retryStrategy sounds very common.

src/client/src/StreamState.c
kateyanurag: changed

src/client/src/ClientState.c
kateyanurag: done

src/client/src/StreamState.c
kateyanurag: done

src/state/include/com/amazonaws/kinesis/video/state/Include.h
kateyanurag: There should not be a problem with producer SDK since producer state machines (client and stream state machines) are defined in PIC. For WebRTC sdk, the signaling state machine is defined in WebRTC sdk itself. And this will be a breaking change. I'm planning to send a new PR for WebRTC sdk to update the signaling state machine.
Do we expect customers to use newer version of PIC and older version of WebRTC? Probably not. If they do, then there will be a problem.

src/utils/src/ExponentialBackoffRetryStrategy.c
kateyanurag: Updated to `checkConfigParameterRange`. I thought of macro but isn't the compiler smart enough to expand this inline? Just trying to avoid a macro.

src/utils/src/ExponentialBackoffRetryStrategy.c
kateyanurag: done

src/state/include/com/amazonaws/kinesis/video/state/Include.h
MushMal: I don't believe any customers are actually using the state machine library as is so updating C producer and the WebRTC will sort it out. However, my main concern is whether you need a separate state transition hook per state? Perhaps, a better approach is to have a single callback function (the hook) specified in the state machine constructor with it's custom data and the first parameter being the state itself (the UINT64 value) that it's being transitioned to. This way, you won't need to change the state machine definitions in the implementation and only provide the new parameter to the create state machine API?

src/state/include/com/amazonaws/kinesis/video/state/Include.h
MushMal: Seems that this is indeed a hook function. Take a look at the other hooks - those are called BlahBlahHookFn or something similar. I really recommend not using the "Error" and "Handler" prefix here at all. Make sure you have a "Hook" prefix as this will be exposed in the public header. Also, I commented below on whether to take a single API instead of per state callback which might not make sense

src/client/src/ClientState.c
MushMal: Formatting makes sense. If you decide to take the hook function at the constructor then you won't need any changes in the state machine definitions.

src/utils/src/ExponentialBackoffRetryStrategy.c
MushMal: Many compilers will not be "smart" enough to inline it. This is not a biggie just macro is more preferred. Be careful with the macro as it uses the params which need to be wrapped in parenthesis and the invocation of the macro should use simple variables and not expressions (which can get re-evaluated)

src/state/include/com/amazonaws/kinesis/video/state/Include.h
hassanctech: @MushMal at a high level let me outline what we're trying to do here.  Today with the state machine there is get next function which determines what the next state is then there is an execute function.  A case which is not handled properly is when the transition from A --> B fails for some reason.  There are two obvious ways to go about it:

1. You can add an error handler function and attach it to each state (it won't be different for each state but it *could* be if needed), this will fire when the transition from A --> B failed for some reason and will react appropriately.  
2. You can introduce an actual error state (which wouldn't require any changes to the state machine impl) clients would simply add this new state, today we do not do this but we could.

We opted to go with (1) because today there is an established flow of states and we felt it would be more difficult to reason about those transitions if we added a new state.  If we were starting from scratch I would have strongly advocated for (2).  In any case, (1) is what we are trying to accomplish in this PR.

With this in mind I do not think this error handler function is a pre hook, it's specific to be invoked when the state transition from A --> B fails for some reason.  And even more specifically today the only thing we're adding here is backoff/jitter before we retry.  Today when you transition A --> A the step state machine function already adds delay. But if you go from A --> B --> A the retry count is reset and this can indeed go in a tight loop.  What we're hoping to do here is have a "global" exponential backoff that's shared across all streams / client.  

Also it is not a state transition function either.  If you're wanting to go from A-->B and `f` is your state transition function then once you execute `f` you are now at state B.  That's not what this is, basically what we're adding is if the transition from A --> B failed for any reason, now lets invoke the error handler.  I understand today we can never technically fail to get from A --> B, we will arrive at B regardless of whether the executefn failed or not (maybe not in all cases).

Having a single callback function in the state machine constructor would not allow for the possibility of different error handler functions for each state.  It's an interesting choice perhaps we can think more about it.

src/client/src/Include_i.h
hassanctech: nit: KVS is an acronym so making the variable name `kVSRetryStrategy` looks weird, how about `kvsRetryStrategy` ?

src/client/src/Client.c
hassanctech: exponentialBackoffRetryStrategyFree will be called which does free the state.

src/state/include/com/amazonaws/kinesis/video/state/Include.h
kateyanurag: Sure I'll change it to `StateTransitionHookFunc`. Replying to the other comment below.

src/client/src/Include_i.h
kateyanurag: Sure. Changed `KVSRetryStrategy` type to `KvsRetryStrategy` and all variables to `kvsRetryStrategy`.
It was a sweeping change. You would see a larger diff now.

src/state/include/com/amazonaws/kinesis/video/state/Include.h
kateyanurag: Thanks for detailed explanation, Hassan! I don't have anything more to add to justify the design choice. @MushMal do you agree with this?
@hassanctech I like the idea of renaming it from `StateErrorHandlerFunc` to `StateTransitionHookFunc`. Transition hook name sounds more apt.

src/utils/src/ExponentialBackoffRetryStrategy.c
kateyanurag: Okay I can change it to a macro (although I'm not super excited about macros :))

kateyanurag:
@hassanctech @MushMal I've pushed a new commit with a fix for broken UT. Travis is green now. Please take a look.

src/client/src/StreamState.c
```diff
@@ -9,36 +9,140 @@
...
+    STATUS retStatus = STATUS_SUCCESS;
+    PKinesisVideoStream pKinesisVideoStream = NULL;
+    PKinesisVideoClient pKinesisVideoClient = NULL;
+    PKvsRetryStrategy pKvsRetryStrategy = NULL;
+
+    pKinesisVideoStream = STREAM_FROM_CUSTOM_DATA(customData);
+    CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
+
+    pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
+    CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
```
hassanctech, line 123: You meant to CHK the client is not null since you already checked the stream above before getting the client out of it.

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -0,0 +1,245 @@
...
+    return retStatus;
+}
+
+STATUS resetExponentialBackoffRetryState(PExponentialBackoffRetryStrategyState pExponentialBackoffRetryStrategyState) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+
+    CHK(pExponentialBackoffRetryStrategyState != NULL, STATUS_NULL_ARG);
+
+    pExponentialBackoffRetryStrategyState->currentRetryCount = 0;
```
hassanctech, line 26: Can this API and `exponentialBackoffRetryStrategyBlockingWait` be called by two separate threads?  If so `lastRetryWaitTime` and `lastRetrySystemTime` are UINT64 which means it is not an atomic operation to set a value (which this API and `exponentialBackoffRetryStrategyBlockingWait` both do, on 32-bit systems. This will result in a torn right.  So we need to be absolutely sure this cannot happen in 2 different threads OR you need to introduce a mutex to protect the setting of these values.

src/client/src/Client.c
```diff
@@ -94,6 +94,64 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
...
+    STATUS retStatus = STATUS_SUCCESS;
+    KVS_RETRY_STRATEGY_TYPE defaultKvsRetryStrategyType = KVS_RETRY_STRATEGY_EXPONENTIAL_BACKOFF_WAIT;
+
+    CHK(pKinesisVideoClient != NULL, STATUS_NULL_ARG);
+
+    // If the callbacks for retry strategy are already set, then use that otherwise
+    // build the client with a default retry strategy.
+    if (pKinesisVideoClient->clientCallbacks.getRetryStrategyFn != NULL) {
+        CHK_STATUS(pKinesisVideoClient->clientCallbacks.getRetryStrategyFn(TO_CLIENT_HANDLE(pKinesisVideoClient)));
+    } else {
```
MushMal, line 109: the else section should be outsourced into a separate function in my opinion - it conflates the setting of the retry callback with the creation and I think those should be separate.

src/client/src/Client.c
```diff
@@ -94,6 +94,64 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
...
+        switch (defaultKvsRetryStrategyType) {
```
MushMal, line 110: This is an invariant and the static code analysis tools should have flagged this.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1963,6 +1978,7 @@ struct __ClientCallbacks {
     LogPrintFunc logPrintFn;
     ClientShutdownFunc clientShutdownFn;
     StreamShutdownFunc streamShutdownFn;
+    GetRetryStrategyFn getRetryStrategyFn;
```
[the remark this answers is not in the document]
MushMal: I think that this shouldn't be in the callbacks at all. Perhaps, the DeviceInfo.ClientInfo structure should be rev-ed to include this strategy information.

src/client/src/ClientState.c
```diff
@@ -9,24 +9,103 @@
...
+        executeReadyClientState,
+        defaultClientStateMachineErrorHandler,
+        INFINITE_RETRY_COUNT_SENTINEL,
+        STATUS_CLIENT_READY_CALLBACK_FAILED
+    }
 };
 
 UINT32 CLIENT_STATE_MACHINE_STATE_COUNT = SIZEOF(CLIENT_STATE_MACHINE_STATES) / SIZEOF(StateMachineState);
 
+STATUS defaultClientStateMachineErrorHandler(UINT64 customData /* customData should be PKinesisVideoClient */) {
```
MushMal, line 79: Naming?

MushMal: Also, should this function be part of the client state machine file?

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -1188,6 +1188,9 @@ typedef UINT64 HANDLE;
         }                                                                                                                                            \
     } while (FALSE)
 
+// Checks if the data provided is in range [low, high]
+#define CHECK_IN_RANGE(data, low, high) ((data) >= (low) && (data) <= (high) ? TRUE : FALSE)
```
MushMal, line 1192: No need for ? TRUE : FALSE as the expression will be evaluated to a boolean

src/state/include/com/amazonaws/kinesis/video/state/Include.h
```diff
@@ -64,6 +64,17 @@ typedef STATUS (*GetNextStateFunc)(UINT64, PUINT64);
...
+/**
+ * Function which gets called for each state before state machine
+ * transitions to a different state. State machine transitioning to
+ * the same state will not result in the transition hook being called
+ *
+ * @param 1 - IN - Custom data passed in
+ *
+ * @return Status of the callback
+ */
+typedef STATUS (*StateTransitionHookFunc)(UINT64);
```
MushMal, line 76: An alternative would be for the ExecuteStateFunc to take a UINT64 which would be the state for which it's called allowing a wrapper API specified in the state machine definition to handle certain logic like retry and then switch/case and call into the state specific execute function.

src/state/src/State.c
```diff
@@ -158,6 +158,12 @@ STATUS stepStateMachine(PStateMachine pStateMachine)
         pStateMachineImpl->context.retryCount = 0;
         pStateMachineImpl->context.time = time;
 
+        // Call the state machine error handler
+        // This call could be a no-op if the state transition is happening for SERVICE_CALL_RESULT_OK code
+        if (pStateMachineImpl->context.pCurrentState->stateTransitionHookFunc != NULL) {
+            CHK_STATUS(pStateMachineImpl->context.pCurrentState->stateTransitionHookFunc(pStateMachineImpl->customData));
```
MushMal, line 164: Well, perhaps the state transition hook is a little more elegant way

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -0,0 +1,245 @@
...
+        if (currentRetryWaitTime > pRetryConfig->maxRetryWaitTime) {
+            currentRetryWaitTime = pRetryConfig->maxRetryWaitTime;
+        }
+    }
+
+    // Note: Do not move the call to getRandomJitter in calculateWaitTime.
+    // This is because we need randomization for wait time after we reach maxRetryWaitTime
+    jitter = getRandomJitter(pRetryState->exponentialBackoffRetryStrategyConfig.jitterFactor);
+    currentRetryWaitTime += jitter;
+    THREAD_SLEEP(currentRetryWaitTime);
```
MushMal, line 208: This is going to be scary. Would be nice to properly comment and document the top-level scenario when this is executed. For example, what scenario this will be used with the values. In case of create stream the wait will be on the thread of execution. The thread could come from the networking thread in C Producer. How long this will be blocked for? Another scenario is the token rotation - in this case the thread is the thread of the new session handlers PutStream API from c producer. WebRTC scenarios will also be two fold - the main execution thread and the network handler/listener thread.

MushMal (commented):
Added some comments to the implementation. The scariest thing is the THREAD_SLEEP which will be called in the state machine implementations. Not sure if there is an easy alternative but I wonder if the better solution is instead of blocking the caller thread is to modify the callAfter value for the state transitions. The value is respected in both WebRTC networking implementation a well as the C Producers networking.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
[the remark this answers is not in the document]
hassanctech: the client info is exactly where I thought this should go as well.

hassanctech:
> Added some comments to the implementation. The scariest thing is the THREAD_SLEEP which will be called in the state machine implementations. Not sure if there is an easy alternative but I wonder if the better solution is instead of blocking the caller thread is to modify the callAfter value for the state transitions. The value is respected in both WebRTC networking implementation a well as the C Producers networking.

Hmm that's a good point all the execute functions set the `callAfter` value and this happens after a call to step the state machine.  I wasn't fully familiar with this before but it looks like the whole point of this is to inform the client when to call next, I suppose it could be abused/ignored and I see we actually save that info in the diagnostics.

src/utils/src/ExponentialBackoffRetryStrategy.c
kateyanurag: Yep. I agree. I did not have a good understanding of the threading model and it makes sense that we should not block the caller thread. I've incorporated this change in the latest commit. Please take a look.
Instead of callAfter, I've directly updated `pStateMachineImpl->context.time`.

src/state/include/com/amazonaws/kinesis/video/state/Include.h
kateyanurag: Using a transition hook seems to me a cleaner approach than what you are suggesting

src/client/src/ClientState.c
kateyanurag: renamed to `defaultClientStateTransitionHook`

src/client/include/com/amazonaws/kinesis/video/client/Include.h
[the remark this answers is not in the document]
kateyanurag: Moved to client info. I've also moved the `RetryStrategy` from client struct to client info struct

hassanctech (dismissed):
ok

hassanctech (approved):
lgtm

PR #144, opened 2021-10-05
Events mkv size bug

jdelapla:
*Issue #, if available:*

*Description of changes:*


kateyanurag:
Change looks right to me but a few tests are failing. Can you please check @jdelapla?

src/client/src/Stream.c
```diff
@@ -2996,9 +2997,25 @@ STATUS packageStreamMetadata(PKinesisVideoStream pKinesisVideoStream, MKV_STREAM
...
+                    CHK_STATUS(mkvgenIncreaseTagsTagSize(tagsStart, simpleSizes));
+                }
+                tagsFound = TRUE;
+                simpleSizes = 0;
+                tagsStart = pBuffer + packagedSize;
+            } else if (pSerializedMetadata->parent != MKV_TREE_SIMPLE) {
+                // if the tag is neither
+                tagsFound = FALSE;
+            }
+
```
hassanctech, line 3011: is this `else if` block needed at all?  The value was already initialized to `FALSE`.

jdelapla: I was trying to handle future cases where items other than TAGS or SIMPLE could be in this linked list, but it may be best to handle that case when it develops. The current design would have misses with this inclusion anyways, such as {TAGS-SIMPLE-TAG-SIMPLE-TAGS}.

Removed it.

hassanctech (approved):
lgtm

PR #148, opened 2021-10-12
There are several missing 'accepted' states from the Stream State mac…

jdelapla:
…hine

*Issue #, if available:*

*Description of changes:*


src/client/src/StreamState.c
```diff
@@ -11,26 +11,27 @@
...
+    {STREAM_STATE_TAG_STREAM, STREAM_STATE_DESCRIBE | STREAM_STATE_CREATE | STREAM_STATE_TAG_STREAM, fromTagStreamState, executeTagStreamState,
+     SERVICE_CALL_MAX_RETRY_COUNT, STATUS_TAG_STREAM_CALL_FAILED},
     {STREAM_STATE_GET_ENDPOINT,
-     STREAM_STATE_STOPPED | STREAM_STATE_DESCRIBE | STREAM_STATE_CREATE | STREAM_STATE_GET_ENDPOINT | STREAM_STATE_TAG_STREAM,
+     STREAM_STATE_STOPPED | STREAM_STATE_DESCRIBE | STREAM_STATE_CREATE | STREAM_STATE_GET_ENDPOINT | STREAM_STATE_TAG_STREAM |
+         STREAM_STATE_PUT_STREAM,
      fromGetEndpointStreamState, executeGetEndpointStreamState, SERVICE_CALL_MAX_RETRY_COUNT, STATUS_GET_STREAMING_ENDPOINT_CALL_FAILED},
-    {STREAM_STATE_GET_TOKEN, STREAM_STATE_STOPPED | STREAM_STATE_GET_ENDPOINT | STREAM_STATE_GET_TOKEN, fromGetTokenStreamState,
-     executeGetTokenStreamState, SERVICE_CALL_MAX_RETRY_COUNT, STATUS_GET_STREAMING_TOKEN_CALL_FAILED},
+    {STREAM_STATE_GET_TOKEN, STREAM_STATE_STOPPED | STREAM_STATE_GET_ENDPOINT | STREAM_STATE_GET_TOKEN | STREAM_STATE_PUT_STREAM,
```
MushMal, line 24: Wondering what's the case when moving from PutStream state to GetToken? Theoretically, there could be a case when a token can expire while going through the states. More importantly, these need to be covered in unit tests

src/client/src/StreamState.c
```diff
@@ -11,26 +11,27 @@
...
+     fromGetTokenStreamState, executeGetTokenStreamState, SERVICE_CALL_MAX_RETRY_COUNT, STATUS_GET_STREAMING_TOKEN_CALL_FAILED},
     {STREAM_STATE_READY, STREAM_STATE_STOPPED | STREAM_STATE_GET_TOKEN | STREAM_STATE_READY | STREAM_STATE_PUT_STREAM | STREAM_STATE_STREAMING,
      fromReadyStreamState, executeReadyStreamState, SERVICE_CALL_MAX_RETRY_COUNT, STATUS_STREAM_READY_CALLBACK_FAILED},
-    {STREAM_STATE_PUT_STREAM, STREAM_STATE_STOPPED | STREAM_STATE_READY | STREAM_STATE_PUT_STREAM, fromPutStreamState, executePutStreamState,
+    {STREAM_STATE_PUT_STREAM, STREAM_STATE_READY | STREAM_STATE_PUT_STREAM, fromPutStreamState, executePutStreamState, INFINITE_RETRY_COUNT_SENTINEL,
```
MushMal, line 28: I thought rhere are cases where you can come to PutStream from StreamStopped state. Do we cycle through describe or others in case of network interruption/disconnect or go directly to put stream?

src/client/src/StreamState.c
```diff
@@ -11,26 +11,27 @@
...
+     STATUS_PUT_STREAM_CALL_FAILED},
+    {STREAM_STATE_STREAMING, STREAM_STATE_PUT_STREAM | STREAM_STATE_STREAMING, fromStreamingStreamState, executeStreamingStreamState,
```
MushMal, line 30: Same question here

src/client/src/StreamState.c
```diff
@@ -11,26 +11,27 @@
...
      INFINITE_RETRY_COUNT_SENTINEL, STATUS_PUT_STREAM_CALL_FAILED},
-    {STREAM_STATE_STREAMING, STREAM_STATE_STOPPED | STREAM_STATE_PUT_STREAM | STREAM_STATE_STREAMING, fromStreamingStreamState,
-     executeStreamingStreamState, INFINITE_RETRY_COUNT_SENTINEL, STATUS_PUT_STREAM_CALL_FAILED},
     {STREAM_STATE_STOPPED,
      STREAM_STATE_STOPPED | STREAM_STATE_CREATE | STREAM_STATE_DESCRIBE | STREAM_STATE_TAG_STREAM | STREAM_STATE_GET_ENDPOINT |
-         STREAM_STATE_GET_TOKEN | STREAM_STATE_READY | STREAM_STATE_PUT_STREAM | STREAM_STATE_STREAMING,
+         STREAM_STATE_GET_TOKEN | STREAM_STATE_READY | STREAM_STATE_PUT_STREAM | STREAM_STATE_STREAMING | STREAM_STATE_NEW,
```
MushMal, line 34: Not sure if we can hit stopped state from the new state

MushMal (commented):
Great to see someone is looking at the state transitions!

src/client/src/StreamState.c
hassanctech: @MushMal according to the `fromNewStreamState` we can go to STOPPED or NEW.

disa6302:
Let us add some details on possible scenarios the new state transitions could occur. I would also like to highlight that we should check other code paths where state transitions might be induced to see if anything specific to new transitions should be handled.

src/client/src/StreamState.c
```diff
@@ -44,7 +45,7 @@ STATUS fromNewStreamState(UINT64 customData, PUINT64 pState)
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     PKinesisVideoStream pKinesisVideoStream = STREAM_FROM_CUSTOM_DATA(customData);
-    UINT64 state;
+    UINT64 state = STREAM_STATE_NEW;
```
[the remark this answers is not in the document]
jdelapla: To prevent undefined behavior. You're right it doesn't functionally change the code but it does follow the same pattern as other 'from' function points and removed the potential for undefined behavior in future revisions.

disa6302 (approved):

PR #149, opened 2021-11-03
Log git hash

disa6302:
*Issue #, if available:*

*Description of changes:*
We do not log the SDK version / git hash. This will be super convenient when debugging issues or when trying to get details from customer on the git version that is being used. When used in producer C, the git hash of producer SDK would be available as an INFO log and when used in WebRTC, the git hash of WebRTC SDK will be available. 


hassanctech (approved):
will be called inside `createStream` so at the start of streaming :).

PR #151, opened 2021-11-19
block tags before first cluster, post tags after

jdelapla:
*Issue #, if available:*

*Description of changes:*


src/client/src/Stream.c
```diff
@@ -900,7 +900,7 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
         CHK_STATUS(mkvgenPackageFrame(pKinesisVideoStream->pMkvGenerator, pFrame, pTrackInfo, pAlloc, &packagedSize, &encodedFrameInfo));
 
         // Package the metadata if specified
-        if (packagedMetadataSize != 0) {
+        if (packagedMetadataSize != 0 && !justStartedStreaming) {
```
hassanctech, line 903: `justStartedStreaming` is never reset so it will always be `FALSE`

hassanctech: Also we should now add a test which actually tests if we block tags before first cluster which is the desired outcome after this change.

disa6302 (approved):

PR #152, opened 2021-11-23
Expose retry count

disa6302:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
+++ b/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
@@ -1520,12 +1520,13 @@ typedef STATUS (*ExecuteRetryStrategyFn)(PRetryStrategy, PUINT64);
  * A generic retry strategy
  */
 struct __KvsRetryStrategy {
-    // Retry strategy type as defined in the above enum
-    KVS_RETRY_STRATEGY_TYPE retryStrategyType;
     // Pointer to metadata/state/details for the retry strategy.
     // The actual data type is abstracted and will be inferred by
     // the RetryHandlerFn
     PRetryStrategy pRetryStrategy;
+
+    // Retry strategy type as defined in the above enum
+    KVS_RETRY_STRATEGY_TYPE retryStrategyType;
 
     // Pointer to the function to create new retry strategy
     CreateRetryStrategyFn createRetryStrategyFn;
@@ -1634,10 +1635,8 @@ typedef struct __ExponentialBackoffRetryStrategyConfig {
     // Factor determining random jitter value.
     // Jitter will be between [0, jitterFactor).
     UINT32  jitterFactor;
-} ExponentialBackoffRetryStrategyConfig;
-typedef ExponentialBackoffRetryStrategyConfig* PExponentialBackoffRetryStrategyConfig;
+} ExponentialBackoffRetryStrategyConfig, *PExponentialBackoffRetryStrategyConfig;
 
-#define TO_EXPONENTIAL_BACKOFF_STATE(ptr)  ((PExponentialBackoffRetryStrategyState)(ptr))
 #define TO_EXPONENTIAL_BACKOFF_CONFIG(ptr) ((PExponentialBackoffRetryStrategyConfig)(ptr))
 
 /**************************************************************************************************
--- a/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
+++ b/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
@@ -1637,7 +1637,21 @@ typedef struct __ExponentialBackoffRetryStrategyConfig {
     UINT32  jitterFactor;
 } ExponentialBackoffRetryStrategyConfig, *PExponentialBackoffRetryStrategyConfig;
 
+#define TO_EXPONENTIAL_BACKOFF_STATE(ptr)  ((PExponentialBackoffRetryStrategyState)(ptr))
 #define TO_EXPONENTIAL_BACKOFF_CONFIG(ptr) ((PExponentialBackoffRetryStrategyConfig)(ptr))
+
+
+typedef struct {
+    ExponentialBackoffRetryStrategyConfig exponentialBackoffRetryStrategyConfig;
+    ExponentialBackoffStatus status;
+    UINT32 currentRetryCount;
+    // The system time at which last retry happened
+    UINT64 lastRetrySystemTime;
+    // The actual wait time for last retry
+    UINT64 lastRetryWaitTime;
+    // Lock to update operations
+    MUTEX retryStrategyLock;
+} ExponentialBackoffRetryStrategyState, *PExponentialBackoffRetryStrategyState;
 
 /**************************************************************************************************
 API usage:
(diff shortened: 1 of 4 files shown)


src/utils/src/Include_i.h
```diff
@@ -270,6 +270,19 @@ typedef struct {
...
+    ExponentialBackoffStatus status;
+    UINT32 currentRetryCount;
+    // The system time at which last retry happened
+    UINT64 lastRetrySystemTime;
+    // The actual wait time for last retry
+    UINT64 lastRetryWaitTime;
+    // Lock to update operations
+    MUTEX retryStrategyLock;
+} ExponentialBackoffRetryStrategyState;
+typedef ExponentialBackoffRetryStrategyState* PExponentialBackoffRetryStrategyState;
```
hassanctech, line 284: You need a `,` between these two.

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,19 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
     return retStatus;
 }
 
+STATUS getExponentialBackoffRetryCount(PRetryStrategy pRetryStrategy, PUINT32 retryCount) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
```
kateyanurag, line 237: Nit: add null check for `pRetryStrategy`

src/utils/src/Include_i.h
hassanctech: I see this is differnet, why not just put it on line 283,
```
} ExponentialBackoffRetryStrategyState, *PExponentialBackoffRetryStrategyState
```

like we do for other structs

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,19 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
...
+    PExponentialBackoffRetryStrategyState pRetryState = NULL;
+    MUTEX_LOCK(pRetryState->retryStrategyLock);
```
hassanctech, line 239: This will crash right.... you just set `pRetryState` to NULL and now you're deref'ing it is this a typo should it be `pRetryStrategy` instead?

kateyanurag: That's a good point. Divya, I think we can pass pRetryStrategy OR typecast pRetryState to KvsRetryStrategy (by making pRetryState as a first element of KvsRetryStrategy struct). Then we should validate KvsRetryStrategy->retryStrategyType if exponential backoff. And then only proceed.

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,20 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
     return retStatus;
 }
 
+STATUS getExponentialBackoffRetryCount(PRetryStrategy pRetryStrategy, PUINT32 retryCount) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PExponentialBackoffRetryStrategyState pRetryState = NULL;
+    CHK_STATUS(pRetryStrategy == NULL, STATUS_NULL_ARG);
+    pRetryState = TO_EXPONENTIAL_BACKOFF_STATE(pRetryStrategy);
+    MUTEX_LOCK(pRetryState->retryStrategyLock);
```
hassanctech, line 241: is it possible that `pRetryState` is `NULL`?

src/utils/src/ExponentialBackoffRetryStrategy.c
disa6302: Done

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,20 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
     return retStatus;
 }
 
+STATUS getExponentialBackoffRetryCount(PRetryStrategy pRetryStrategy, PUINT32 retryCount) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PExponentialBackoffRetryStrategyState pRetryState = NULL;
+    pRetryState = TO_EXPONENTIAL_BACKOFF_STATE(pRetryStrategy);
+    CHK_STATUS(pRetryStrategy != NULL && pRetryState != NULL, STATUS_NULL_ARG);
```
hassanctech: We should use CHK here not CHK_STATUS, CHK evaluates the condition in an `if` but CHK_STATUS applies `STATUS_FAILED` its meant to check value that's a STATUS but here it's a boolean.

src/utils/src/ExponentialBackoffRetryStrategy.c
disa6302: It is, I am doing a NULL check to avoid any dereferencing.

src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
```diff
@@ -1624,19 +1637,6 @@ typedef struct __ExponentialBackoffRetryStrategyConfig {
 } ExponentialBackoffRetryStrategyConfig;
 typedef ExponentialBackoffRetryStrategyConfig* PExponentialBackoffRetryStrategyConfig;
```
hassanctech, line 1638: I know its not from your PR but could you please fix this too to be consistent with our other typedefs

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,20 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
     return retStatus;
 }
 
+STATUS getExponentialBackoffRetryCount(PRetryStrategy pRetryStrategy, PUINT32 retryCount) {
```
hassanctech, line 235: How/where will this be used?  The question is really do we need to pass a `PRetryStrategy` to this function it looks like it doesn't actually use anything from it except for the retry state, can we just pass it the `PRetryState` instead?  Perhaps there is a good reason to have it, would like to understand.

disa6302: PRetryState is supposed to be an internal structure. This PR moves the struct to `Include_i.h`

src/utils/src/ExponentialBackoffRetryStrategy.c
```diff
@@ -232,6 +232,29 @@ STATUS getExponentialBackoffRetryStrategyWaitTime(PRetryStrategy pRetryStrategy,
...
 
+STATUS getExponentialBackoffRetryCount(PRetryStrategy pRetryStrategy, PUINT32 pRetryCount) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PKvsRetryStrategy pKvsRetryStrategy = NULL;
+    PExponentialBackoffRetryStrategyState pRetryState = NULL;
+
+    CHK(pRetryStrategy != NULL && pRetryCount != NULL, STATUS_NULL_ARG);
+    pKvsRetryStrategy = (PKvsRetryStrategy)(pRetryStrategy);
+    pRetryState = TO_EXPONENTIAL_BACKOFF_STATE(pRetryStrategy);
```
kateyanurag, line 243: Move this line after the CHK on line 247

kateyanurag (approved):

hassanctech (approved):
lgtm

PR #158, opened 2021-12-03
Create retry strategy after fixing up device info

kateyanurag:
*Issue #, if available:*

*Description of changes:*
Crate retry strategy after fixing up device info. `fixupDeviceInfo` overwrites deviceInfo.clientInfo and hence we were losing the retry strategy callbacks which are set in configureClientWithRetryStrategy.

### DIFF
--- a/src/client/src/Client.c
+++ b/src/client/src/Client.c
@@ -221,13 +221,13 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
     // Copy the structures in their entirety
     MEMCPY(&pKinesisVideoClient->clientCallbacks, pClientCallbacks, SIZEOF(ClientCallbacks));
 
-    CHK_STATUS(configureClientWithRetryStrategy(pKinesisVideoClient));
-
     // Fix-up the defaults if needed
     // IMPORTANT!!! The calloc allocator will zero the memory which will also act as a
     // sentinel value in case of an earlier version of the structure
     // is used and the remaining fields are not copied
     fixupDeviceInfo(&pKinesisVideoClient->deviceInfo, pDeviceInfo);
+
+    CHK_STATUS(configureClientWithRetryStrategy(pKinesisVideoClient));
 
     // Fix-up the name of the device if not specified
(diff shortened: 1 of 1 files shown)


src/client/src/Client.c
```diff
@@ -221,14 +221,14 @@ STATUS createKinesisVideoClient(PDeviceInfo pDeviceInfo, PClientCallbacks pClien
...
-    CHK_STATUS(configureClientWithRetryStrategy(pKinesisVideoClient));
-
     // Fix-up the defaults if needed
     // IMPORTANT!!! The calloc allocator will zero the memory which will also act as a
     // sentinel value in case of an earlier version of the structure
     // is used and the remaining fields are not copied
     fixupDeviceInfo(&pKinesisVideoClient->deviceInfo, pDeviceInfo);
 
+    CHK_STATUS(configureClientWithRetryStrategy(pKinesisVideoClient));
+
```
hassanctech, line 231: `fixupDeviceInfo` in turn calls `fixupClientInfo` whose job is to just set up the defaults.  I don't see how this would cause the retry strategy callbacks to be lost?  Even if it were the case you need to add logic in `fixupClientInfo` to see if they're set so it doesn't try to modify them and/or set it to the defaults there like we do for other attributes in clientInfo/deviceInfo.

kateyanurag: In `fixupClientInfo` I'm already copying retry strategy callbacks. 
https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/bug-fix/src/client/src/InputValidator.c#L311
But these callbacks are optional. So if they are not present, NULL data will be copied. So we should be calling `configureClientWithRetryStrategy` after `fixupClientInfo` to make sure that if no callbacks are provided, then use our own. This way I'm able to inject callbacks in unit tests. I couldn't find a better way. And also like the `fixup` logic. It confused me a lot.

hassanctech: Ok we can revisit later but maybe I'm missing something but if you call it *before* it won't be null anymore right so then can't you just do a null check if its not null then don't overwrite it?

hassanctech (approved):
lgtm

PR #161, opened 2021-12-15
Retry count Metric

disa6302:
*Issue #, if available:*

*Description of changes:*
- Add retry count metric for every stream
- Add client level average retry count metric for all streams 

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -710,13 +710,13 @@ STATUS logStreamMetric(PKinesisVideoStream pKinesisVideoStream)
     DLOGD("\tAverage Data Plane API latency (ms): %" PRIu64 " ", streamMetrics.dataApiCallLatency / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
 
     // V2 stream information
-    DLOGD("\tCurrent elementary frame rate (fps): %f ", streamMetrics.elementaryFrameRate);
+    DLOGD("\tCurrent elementary frame rate (fps): %lf ", streamMetrics.elementaryFrameRate);
 
     // V3 stream information
     DLOGD("\tAPI Call Retry Count : %lu", streamMetrics.streamApiCallRetryCount);
 
     // V1 client information
-    DLOGD("\tTotal elementary frame rate (fps): %f ", clientMetrics.totalElementaryFrameRate);
+    DLOGD("\tTotal elementary frame rate (fps): %lf ", clientMetrics.totalElementaryFrameRate);
 
     // V2 client information
     DLOGD("\tAverage API call retry count for client: %lf", clientMetrics.clientAvgApiCallRetryCount);
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -713,7 +713,7 @@ STATUS logStreamMetric(PKinesisVideoStream pKinesisVideoStream)
     DLOGD("\tCurrent elementary frame rate (fps): %f ", streamMetrics.elementaryFrameRate);
 
     // V3 stream information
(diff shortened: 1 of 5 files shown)


src/client/src/StreamState.c
```diff
@@ -138,6 +139,15 @@ STATUS defaultStreamStateTransitionHook(
             pKvsRetryStrategy->pRetryStrategy != NULL &&
             pKvsRetryStrategyCallbacks->executeRetryStrategyFn != NULL, STATUS_SUCCESS);
 
+    if(pKvsRetryStrategyCallbacks->getCurrentRetryAttemptNumberFn != NULL) {
+        if((countStatus = pKvsRetryStrategyCallbacks->getCurrentRetryAttemptNumberFn(pKvsRetryStrategy, &pKinesisVideoStream->diagnostics.streamApiCallRetryCount)) != STATUS_SUCCESS) {
+            DLOGW("Failed to get retry count. Error code: %08x", countStatus);
+        }
+        else {
+            DLOGD("Client state machine retry count: %llu", pKinesisVideoStream->diagnostics.streamApiCallRetryCount);
```
hassanctech, line 147: UINT32 format specifier should be just `%lu`

hassanctech (commented):
Some API results include caching.  Is this metric explicit in including actual API calls or just code paths that could make the API calls and if it's the latter then do we need to make some changes to make sure this is only incremented when the actual backend APIs are called instead of when it's possible a cached value is retrieved?

disa6302:
> Some API results include caching. Is this metric explicit in including actual API calls or just code paths that could make the API calls and if it's the latter then do we need to make some changes to make sure this is only incremented when the actual backend APIs are called instead of when it's possible a cached value is retrieved?

It should be. This metric just reads what gets updated in exponential backoff. So if the count gets incremented in exponential backoff, the metric would reflect the same.

hassanctech (commented):
Also, where is `streamApiCallRetryCount` actually incremented?

src/client/src/StreamState.c
disa6302: Oops. Fixed in new commit.

disa6302:
> Also, where is `streamApiCallRetryCount` actually incremented?

This is where it gets incremented: https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/d10936c70ef00941226f431b9bda46eca5aa2f0a/src/utils/src/ExponentialBackoffRetryStrategy.c#L271 and `streamApiCallRetryCount` just gets updated with the value through the `getCurrentRetryAttemptNumberFn` callback.

src/client/src/Stream.c
```diff
@@ -712,8 +712,14 @@ STATUS logStreamMetric(PKinesisVideoStream pKinesisVideoStream)
     // V2 stream information
     DLOGD("\tCurrent elementary frame rate (fps): %f ", streamMetrics.elementaryFrameRate);
 
+    // V3 stream information
+    DLOGD("\tAPI Call Retry Count : %llu", streamMetrics.streamApiCallRetryCount);
```
hassanctech, line 716: This one too `%lu`

src/client/src/Stream.c
```diff
@@ -712,8 +712,14 @@ STATUS logStreamMetric(PKinesisVideoStream pKinesisVideoStream)
...
+
     // V1 client information
     DLOGD("\tTotal elementary frame rate (fps): %f ", clientMetrics.totalElementaryFrameRate);
```
hassanctech, line 719: `%lf` for clientMetrics.totalElementaryFrameRate as well since its also `DOUBLE`

hassanctech (commented):
lgtm let's just wait for travis to finish

hassanctech (approved):
CI is clean, LGTM!

PR #162, opened 2022-01-05
initial changes for supporting setting / retrieving current time offset

hassanctech:
*Issue #, if available:*

*Description of changes:*


hassanctech:
Closing because this is incomplete, and was intended as a stepping stone to having automatic clock skew correction in Producer C.  A complication that arised there is the auth path vs control plane calls have different structs and different callbacks for the curl transfer and in certain paths the curl callbacks to retrieve the http headers which contain the time information are not even implemented.  This is a larger work item.

PR #176, opened 2022-04-22
fix deadlock

hassanctech:
*Issue #, if available:* https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/issues/299

*Description of changes:*
See Issue, fix deadlock

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -748,6 +748,7 @@ STATUS putFrame(PKinesisVideoStream pKinesisVideoStream, PFrame pFrame)
     UINT64 windowDuration, currentDuration;
     PTrackInfo pTrackInfo = NULL;
     PSerializedMetadata pSerializedMetadata = NULL;
+    PFrameOrderCoordinator pFrameOrderCoordinator = pKinesisVideoStream->pFrameOrderCoordinator;
 
     CHK(pKinesisVideoStream != NULL && pFrame != NULL, STATUS_NULL_ARG);
     pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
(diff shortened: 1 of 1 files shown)


MushMal (dismissed):
wow

MushMal (dismissed):

src/client/src/Client.c
```diff
@@ -50,6 +50,7 @@ STATUS checkIntermittentProducerCallback(UINT32 timerId, UINT64 currentTime, UIN
     pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.lock);
 
     if (STATUS_SUCCEEDED(retStatus)) {
+        pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoClient->base.putFrameLock);
```
hassanctech, line 53: Note this is a client level lock. So it is what some may call a "giant lock", but for correctness (avoiding deadlock) in the present state of the code, this is necessary because in the intermittent producer callback we actually need to acquire this lock but if it's a stream level lock we wouldn't have access to it until after we iterate over the streams (which requires we in advance have the `streamListLock`) and if we have that then we can't *first* acquire this putFrame lock it needs to be acquired *before* acquiring the `streamListLock` and it's impossible to do this.

disa6302:
@hassanctech ,

Can you direct the PR to develop?
Ideally, Travis builds would not work anymore and master still has Travis. We want to run GHA on the PR.

hassanctech:
@disa6302 It was easier to just create a new PR:  https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/177

disa6302:
Thank you @hassanctech !

hassanctech:
Looks like CI is all good on https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/177

MushMal:
Regarding the logging... As it turns out, it reaches out and locks a higher-level lock in order to enumerate the streams. As you've noted, this is the root cause of the deadlock. Unfortunately, this becomes pretty messy. The logging was done in the put frame before as we had no source of liveness. Perhaps now since there is a timer queue for the intermittent producer, we can remove the logging code from the putframe and put it in the timer queue. This would eliminate the need to relock the stream and the client and get rid of the logic that calculates when the next time to log would be? The change would be pretty localized. It would also be nice to start thinking about getting rid of the entire client object and only leave stream as the top level object.

hassanctech:
+1 for moving the `logStreamMetric` stuff outside putFrame and into it's own timer queue call would be ideal, but for now we should merge: https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/177, I think it will take some careful consideration to actually do that work, but this was my suggestion as well as the long term goal to simplify all of this.

disa6302:
Closing this PR in lieu of #177

PR #181, opened 2022-11-29
Fail create call with flag 1 4 1

hassanctech:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/client/src/Stream.h
+++ b/src/client/src/Stream.h
@@ -413,7 +413,7 @@ struct __KinesisVideoStream {
     // Whether to set the non-key frames as "SKIP" when we have a not-yet-completed errored fragment
     BOOL skipNonKeyFrames;
 
-    BOOL allowToCreate;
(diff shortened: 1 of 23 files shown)


src/client/src/Stream.h
```diff
@@ -413,6 +413,8 @@ struct __KinesisVideoStream {
     // Whether to set the non-key frames as "SKIP" when we have a not-yet-completed errored fragment
     BOOL skipNonKeyFrames;
 
+    BOOL allowToCreate;
```
MushMal, line 416: Perhaps for consistency call it the same as in stream caps = allowStreamCreation?

MushMal (approved):

disa6302 (dismissed):

src/client/src/Stream.h
hassanctech: Agree, done.

disa6302 (approved):

PR #186, opened 2023-03-29
Add completion timeout configurability

disa6302:
*Issue #, if available:*

*Description of changes:*
- Add configurability for curl call connection and completion timeouts. 
- Fix compiler warning for sprintf in several unit tests
- Add unit test for version (client info)

### DIFF
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -322,9 +322,15 @@ extern "C" {
 #define MIN_VIEW_BUFFER_DURATION (MAX(MIN_BUFFER_DURATION_IN_SECONDS * HUNDREDS_OF_NANOS_IN_A_SECOND, MIN_CONTENT_VIEW_BUFFER_DURATION))
 
 /**
- * Service call default timeout - 5 seconds
+ * Service call default connection timeout - 5 seconds
  */
-#define SERVICE_CALL_DEFAULT_TIMEOUT (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+#define SERVICE_CALL_DEFAULT_CONNECTION_TIMEOUT (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+
+/**
+ * Service call default completion timeout - 10 seconds. Needs to be more than connection timeout
+ */
+#define SERVICE_CALL_DEFAULT_TIMEOUT (10 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+
 
 /**
  * Service call infinite timeout for streaming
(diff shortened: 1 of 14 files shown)


src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1148,6 +1154,8 @@ typedef struct __ClientInfo {
     // Version of the struct
     UINT32 version;
 
+    // ------------------------------- V0 compat ----------------------
```
hassanctech, line 1157: The previous compat lines were correct.  They indicate that everything above the comment line is compatible with that version.  V0 included up until the log metric.  See this commit is where `ClientInfo` was introduced and in the same commit the `CLIENT_INFO_CURRENT_VERSION` was set to `0`.

MushMal (dismissed):
Other than @hassanctech comment, I would recommend adding a couple of version-specific tests.

src/client/include/com/amazonaws/kinesis/video/client/Include.h
disa6302: Fixed

disa6302:
> Other than @hassanctech comment, I would recommend adding a couple of version-specific tests.

Done

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -1484,6 +1495,9 @@ struct __ServiceCallContext {
 
     // Authentication info
     PAuthInfo pAuthInfo;
+
+    // -------- V1 compat --------
```
hassanctech, line 1499: This should be V0 compat, marking the end of V0, to match the format in our other versioned structs.  See for example `ClientInfo` and `StreamCaps`

hassanctech (approved):
LGTM!

jdelapla (changes requested):
With a core struct being changed we also need the associated JNI change before we take this in.

src/client/src/InputValidator.c
```diff
@@ -326,6 +333,23 @@ VOID fixupClientInfo(PClientInfo pClientInfo, PClientInfo pOrigClientInfo)
...
+        pClientInfo->serviceCallCompletionTimeout = SERVICE_CALL_DEFAULT_TIMEOUT;
+    }
+
+    if(pClientInfo->serviceCallConnectionTimeout == 0 || !IS_VALID_TIMESTAMP(pClientInfo->serviceCallConnectionTimeout)) {
+        // Although curl sets default if the value is 0, the default is 300 seconds (5 minutes) which is very high
+        DLOGW("Connection timeout should be non zero...setting to default");
+        pClientInfo->serviceCallConnectionTimeout = SERVICE_CALL_DEFAULT_CONNECTION_TIMEOUT;
+    }
+
+    if(pClientInfo->serviceCallCompletionTimeout == 0 || !IS_VALID_TIMESTAMP(pClientInfo->serviceCallCompletionTimeout)) {
```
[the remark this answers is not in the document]
jdelapla: Nevermind, my eyes glazed over the completion != connection

hassanctech:
> With a core struct being changed we also need the associated JNI change before we take this in.

I don't think so the struct is versioned, so it retains backwards compat with older versions of the struct, if it encounters older versions (which it would if invoked from what is in JNI) then it would use the defaults.

disa6302:
> > With a core struct being changed we also need the associated JNI change before we take this in.
> 
> I don't think so the struct is versioned, so it retains backwards compat with older versions of the struct, if it encounters older versions (which it would if invoked from what is in JNI) then it would use the defaults.

This is true. But I think @jdelapla was trying to achieve feature parity between JNI and Java.

jdelapla (approved):

PR #187, opened 2023-07-01
[ENHANCEMENT] Use new log level

disa6302:
*Issue #, if available:*

*Description of changes:*

What was changed?

Introduced a new log level `PROFILE` which would be used to collect execution time related information for different parts of the code. Also modified the existing log string timestamp to include a millisecond portion for more precision.

Why was it changed?

The change was made to ensure that PROFILE log level related logs get redirected to another log file to allow customers to share it as a report. The functionality is made generic to filter any log level, but the intent began with directing PROFILE logs to another file

How was it changed?

- Added a new log level to allow time profiling different sections of the code
- Added a file logger functionality that allows directing a specific log level into a different file. When using the new API, there will be two log file types that would be generated: one that has the filter specific logs and the other would have other allowed log levels
- Added a new API to generate log timestamps with millisecond precision

Testing:

- Updated existing unit tests and added new unit tests for the new public API
- Tested using the producer C SDK sample with different supported parameters in the existing createFileLogger() and createFileLoggerWithFilter APIs

** CI Changes: **

- Updated all images to ubuntu-latest since 18.04 is not available in GHA anymore.
- Removed GCC 4.4 specific build since ubuntu-latest does not have GCC in the repository. Need to figure out how to test this version on other ubuntu versions given 18.04 is not available in GHA hosted runners: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners


src/utils/src/Time.c
```diff
@@ -3,6 +3,70 @@
...
+
+    timeinfo = GMTIME(&(nowTime.tv_sec));
+    millisecondVal = nowTime.tv_nsec / DEFAULT_TIME_UNIT_IN_NANOS;  // Convert nanoseconds to milliseconds
+
+#endif
+    CHK(pDestBuffer != NULL, STATUS_NULL_ARG);
+    CHK(STRNLEN(formatStr, MAX_TIMESTAMP_FORMAT_STR_LEN + 1) <= MAX_TIMESTAMP_FORMAT_STR_LEN, STATUS_MAX_TIMESTAMP_FORMAT_STR_LEN_EXCEEDED);
+
+    formattedStrLen = 0;
+    *pFormattedStrLen = 0;
```
hassanctech, line 55: Can we add a null check / bail out if `pFormattedStrLen` is NULL?  Especially since this is public API we cannot depend on clients to sanitize input.

I realize it's missing from `generateTimestampStr` function as well.

src/utils/src/Time.c
```diff
@@ -3,6 +3,70 @@
...
+
+#else
+    struct timespec nowTime;
+    clock_gettime(CLOCK_REALTIME, &nowTime);
+
+    timeinfo = GMTIME(&(nowTime.tv_sec));
+    millisecondVal = nowTime.tv_nsec / DEFAULT_TIME_UNIT_IN_NANOS;  // Convert nanoseconds to milliseconds
+
+#endif
+    CHK(pDestBuffer != NULL, STATUS_NULL_ARG);
```
hassanctech, line 51: Let's move this check to top of function after the variable declares, similar to where it is in `generateTimestampStr`, no point in doing those calculations if the dest buffer is NULL.

src/utils/src/Logger.c
```diff
@@ -38,15 +40,15 @@ VOID addLogMetadata(PCHAR buffer, UINT32 bufferLen, PCHAR fmt, UINT32 logLevel)
 #endif
 
     /* if something fails in getting time, still print the log, just without timestamp */
-    retStatus = generateTimestampStr(globalGetTime(), "%Y-%m-%d %H:%M:%S ", timeString, (UINT32) ARRAY_SIZE(timeString), &timeStrLen);
+    retStatus = generateTimestampStrInMilliseconds("%Y-%m-%d %H:%M:%S", timeString, (UINT32) ARRAY_SIZE(timeString), &timeStrLen);
```
hassanctech, line 43: We still need a space here after the `%S`, right? 

So 
```
"%Y-%m-%d %H:%M:%S "
```

Instead of:

```
"%Y-%m-%d %H:%M:%S"
```

src/utils/src/Time.c
disa6302: Added.

src/utils/src/Time.c
disa6302: Oops. It was at the top and I was doing some code changes to fix the builds. Might have got rearranged at the time. Fixed it

src/utils/src/Logger.c
disa6302: Nope. We dont. The format for this function is: "%Y-%m-%d %H:%M:%S.ssssss ". So the space is added after the millisecond portion.

hassanctech: I'm not following, according to the function comments for `generateTimestampStrInMilliseconds` the first parameter is 
```
 * @PCHAR   - IN - timestamp format string
```

The string passed in as the first argument is `%Y-%m-%d %H:%M:%S`, so that should be the timestamp format string.  But you're saying instead the format is:  `%Y-%m-%d %H:%M:%S.ssssss `?  I am confused, where is this latter format coming from?

disa6302: https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/187/files#diff-a54734ddfbb2b0634faf12eae8713f0162689aca80dbe80ee8f53b4a3799cee1R57-R63

The latter is appended inside `generateTimestampStrInMilliseconds` function.

src/utils/src/Time.c
```diff
@@ -3,12 +3,76 @@
...
+    millisecondVal = nowTime.tv_nsec / DEFAULT_TIME_UNIT_IN_NANOS;  // Convert nanoseconds to milliseconds
+
+#endif
+    formattedStrLen = 0;
+    *pFormattedStrLen = 0;
+
+    formattedStrLen = (UINT32) STRFTIME(pDestBuffer, destBufferLen - MAX_MILLISECOND_PORTION_LENGTH, formatStr, timeinfo);
+    CHK(formattedStrLen != 0, STATUS_STRFTIME_FALIED);
+    // Total length is 8 plus terminating null character. Generated string would have utmost size - 1. Hence need to add 1 to max length
+    SNPRINTF(pDestBuffer + formattedStrLen, MAX_MILLISECOND_PORTION_LENGTH + 1, ".%06d ", millisecondVal);
```
hassanctech, line 60: I see ok so no matter what format str is passed in you will add a `.ssssss` to it.  

Can you make that clear in the API docs or comments, because this function is actually appending things to the user passed in format string, where I would have expected if the user can specify a string they would be required to specify the ms component as well and we would format our output to match what the user specified.  I would prefer we didn't append to user provided format string but I think it's fine in this case as long as it's clear to the user of this function that is what is going to happen to their provided string.  I can see why this is the case since the `tm` struct does not seem to contain ms data so I suppose you cannot use it with `strftime`.

src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
```diff
@@ -995,8 +995,19 @@ BOOL checkBufferValues(PVOID, BYTE, SIZE_T);
  */
 PUBLIC_API STATUS generateTimestampStr(UINT64, PCHAR, PCHAR, UINT32, PUINT32);
 
+/**
+ * @PCHAR   - IN - timestamp format string
```
hassanctech, line 999: Since this is a Public API can we provide sample usage here so it's clear that the user isn't expected to pass in the milliseconds part as part of the timestamp format string, because from the name of the function it is not clear that the user of this API is not expected to pass in the milliseconds.

src/utils/src/Logger.c
hassanctech: I've made comments elsewhere, it is not clear from the API definition that you actually do not expect the user to pass in a format string including milliseconds and that you manually append that part yourself.

src/utils/src/Time.c
hassanctech: Also since this is a public API we need to make sure it works for *any* format str the customer passes and if it doesn't to make sure we state what isn't support.  Since you're appending the ".%06d" to the end, for example if the customer only wanted minute precision and specified that as the format string, i.e. `"%Y-%m-%d %H:%M"` this this will not work you'll get a time like:  "2023-07-11 16:26.123456" in which case the meaning is `.123456` of a minute but the value printed is the millisecond so this will be incorrect.

src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
disa6302: Fair. I fixed the documentation for this function and `generateTimestampStr`. Provided a sample and input expectations.

hassanctech (approved):
lgtm

PR #196, opened 2023-07-13
Add threadsafe blocking queue

jdelapla:
*Description of changes:*
Added an implementation and testing for a thread safe blocking queue. Added a new constructor for a semaphore that allows it to behave as a signaling semaphore without adding a 200ms to the destructor.

### DIFF
--- a/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
+++ b/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
@@ -1798,6 +1798,56 @@ PUBLIC_API STATUS exponentialBackoffRetryStrategyFree(PKvsRetryStrategy);
  */
 PUBLIC_API STATUS computePower(UINT64, UINT64, PUINT64);
 
+//////////////////////////////////////////////////////////////////////////////////////////////////////
+// Threadsafe Blocking Queue APIs
+//////////////////////////////////////////////////////////////////////////////////////////////////////
+
+typedef struct {
+    PStackQueue queue;
+    MUTEX       mutex;
+    SEMAPHORE_HANDLE semaphore;
+} SafeBlockingQueue, *PSafeBlockingQueue;
+
+/**
+ * Create a new thread safe blocking queue
+ */
+PUBLIC_API STATUS safeBlockingQueueCreate(PSafeBlockingQueue*);
+
+/**
+ * Frees and de-allocates the thread safe blocking queue
+ */
+PUBLIC_API STATUS safeBlockingQueueFree(PSafeBlockingQueue);
+
+/**
+ * Clears and de-allocates all the items
+ */
+PUBLIC_API STATUS safeBlockingQueueClear(PSafeBlockingQueue, BOOL);
+
+/**
+ * Gets the number of items in the stack/queue
+ */
+PUBLIC_API STATUS safeBlockingQueueGetCount(PSafeBlockingQueue, PUINT32);
+
+/**
+ * Removes the item at the given item
+ */
+PUBLIC_API STATUS safeBlockingQueueRemoveItem(PSafeBlockingQueue, UINT64);
+
+/**
+ * Whether the thread safe blocking queue is empty
+ */
+PUBLIC_API STATUS safeBlockingQueueIsEmpty(PSafeBlockingQueue, PBOOL);
+
+/**
+ * Enqueues an item in the queue
+ */
+PUBLIC_API STATUS safeBlockingQueueEnqueue(PSafeBlockingQueue, UINT64);
+
+/**
+ * Dequeues an item from the queue
+ */
+PUBLIC_API STATUS safeBlockingQueueDequeue(PSafeBlockingQueue, PUINT64);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
+++ b/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
@@ -28,7 +28,7 @@ extern "C" {
  * EMA (Exponential Moving Average) alpha value and 1-alpha value - over appx 20 samples
  */
 #define EMA_ALPHA_VALUE           ((DOUBLE) 0.05)
-#define ONE_MINUS_EMA_ALPHA_VALUE ((DOUBLE)(1 - EMA_ALPHA_VALUE))
+#define ONE_MINUS_EMA_ALPHA_VALUE ((DOUBLE) (1 - EMA_ALPHA_VALUE))
 
 /**
  * Calculates the EMA (Exponential Moving Average) accumulator value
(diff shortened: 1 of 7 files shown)


disa6302:
Whenever this PR is ready, can we wait to merge this PR: https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/193 so that develop is in sync with the main branch?

jdelapla:
> Whenever this PR is ready, can we wait to merge this PR: #193 so that develop is in sync with the main branch?

Sure thing

src/utils/src/ThreadsafeBlockingQueue.c
```diff
@@ -0,0 +1,214 @@
...
+    // are no longer directly interacting with the queue
+    //
+    // free semaphore, this will unblock any threads blocking on said semaphore,
+    // and then exit from the terminate flag
+    //
+    // unlock mutex
+    ATOMIC_STORE_BOOL(&pSafeQueue->terminate, TRUE);
+
+    MUTEX_LOCK(pSafeQueue->mutex);
+    CHK_STATUS(semaphoreFree(&(pSafeQueue->semaphore)));
```
MushMal, line 51: This will leak if the semaphore fails to free. Perhaps remembering the status code of the operation and/or DLOGWing it?

MushMal: I meant will leak the mutex and the queue

src/utils/src/ThreadsafeBlockingQueue.c
```diff
@@ -0,0 +1,214 @@
...
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
+
+    CHK(pSafeQueue != NULL, STATUS_NULL_ARG);
+    CHK(!ATOMIC_LOAD_BOOL(&pSafeQueue->terminate), STATUS_INVALID_OPERATION);
+
+    // 0 timeout semaphore acquire, eventually getting a timeout, this is done to clear
+    // the counting semaphore
+    while ((semaphoreAcquire(pSafeQueue->semaphore, 0) == STATUS_SUCCESS) && !ATOMIC_LOAD_BOOL(&pSafeQueue->terminate))
```
MushMal, line 77: This will peg the CPU core to 100%. There needs to be a short delay to create spinlock type of thing - something like 5-30milliseconds.

src/utils/src/ThreadsafeBlockingQueue.c
```diff
@@ -0,0 +1,214 @@
...
+        MUTEX_UNLOCK(pSafeQueue->mutex);
+    }
+
+    return retStatus;
+}
+
+/**
+ * Whether the thread safe blocking queue is empty
+ */
+STATUS safeBlockingQueueIsEmpty(PSafeBlockingQueue pSafeQueue, PBOOL pIsEmpty)
```
MushMal, line 130: Implement this as a call to get count and return whether the count is 0

src/utils/src/ThreadsafeBlockingQueue.c
```diff
@@ -0,0 +1,214 @@
...
+    MUTEX_LOCK(pSafeQueue->mutex);
+    locked = TRUE;
+    CHK(!ATOMIC_LOAD_BOOL(&pSafeQueue->terminate), STATUS_INVALID_OPERATION);
+
+    CHK_STATUS(stackQueueEnqueue(pSafeQueue->queue, item));
+
+    MUTEX_UNLOCK(pSafeQueue->mutex);
+    locked = FALSE;
+
+    CHK_STATUS(semaphoreRelease(pSafeQueue->semaphore));
```
MushMal, line 175: Perhaps I am not fully understanding the logic of the blocking queue. I would imagine the semaphore is used to ensure the queue waits until the items are all gone before freeing? If so, wouldn't the semaphore permits be acquired when an item gets enqueued and released when an item gets dequeued?

src/utils/tst/ThreadsafeBlockingQueue.cpp
```diff
@@ -0,0 +1,153 @@
...
+    //make it even
+    totalThreads -= totalThreads%2;
+    EXPECT_EQ(STATUS_SUCCESS, safeBlockingQueueCreate(&pSafeQueue));
+    for(UINT64 i = 0; i < totalThreads/2; i++) {
+        THREAD_CREATE(&threads[threadCount++], readingThread, pSafeQueue);
+    }
+    for(UINT64 i = 0; i < totalThreads/2; i++) {
+        THREAD_CREATE(&threads[threadCount++], writingThread, pSafeQueue);
+    }
+    for(UINT64 i = 0; i < totalThreads; i++) {
```
MushMal, line 129: If I understand the logic correctly then a good test would be to move the attach after the safeBLockingQueueFree call

src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
```diff
@@ -1280,6 +1283,16 @@ typedef SEMAPHORE_HANDLE* PSEMAPHORE_HANDLE;
...
 
+/**
+ * Create a semaphore object that starts with 0 count
+ *
+ * @param - UINT32 - IN - The permit count
+ * @param - PSEMAPHORE_HANDLE - OUT - Semaphore handle
+ *
+ * @return  - STATUS code of the execution
+ */
+PUBLIC_API STATUS semaphoreEmptyCreate(UINT32, PSEMAPHORE_HANDLE);
```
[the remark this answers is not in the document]
jdelapla: If this was an initial commit of the library I would agree, but changing the constructor of such a core object now would deprecate a lot of code.

src/utils/src/Semaphore.c
```diff
@@ -88,7 +113,12 @@ STATUS semaphoreCreateInternal(UINT32 maxPermits, PSemaphore* ppSemaphore)
     CHK(maxPermits > 0, STATUS_INVALID_ARG);
     CHK(NULL != (pSemaphore = (PSemaphore) MEMCALLOC(1, SIZEOF(Semaphore))), STATUS_NOT_ENOUGH_MEMORY);
     pSemaphore->maxPermitCount = maxPermits;
-    ATOMIC_STORE(&pSemaphore->permitCount, maxPermits);
+    pSemaphore->signalSemaphore = empty;
```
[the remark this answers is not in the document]
jdelapla: For the same reasoning as above I chose against it. I agree that parameter changes could work fine, but this option allows us to avoid introducing a backwards compatibility issue for a commonly used library across many repos.

src/utils/src/ThreadsafeBlockingQueue.c
jdelapla: Missed the lock problem since semaphoreFree() can only fail if the semaphore_handle is NULL, which would mean the constructor failed.

With that in mind I'm removing the CHK_STATUS on both of the Free() calls here. The destructors of the utility objects only fail (and should only fail) in the event of a null object, so the check doesn't offer anything.

src/utils/src/ThreadsafeBlockingQueue.c
jdelapla: Not quite. In the event the semaphore count is reduced to 0 this operation won't return to STATUS_SUCCESS, it'll return some variation of STATUS_TIMEOUT, which is what we want. A short delay just slows down this finite operation.

src/utils/src/ThreadsafeBlockingQueue.c
jdelapla: The semaphore acts as a count of items in the queue, and forces operations to block when the queue is empty. A signal with a count of permissions of quick access.

src/utils/src/ThreadsafeBlockingQueue.c
jdelapla: The implementation for ThreadsafeBlockingQueue directly maps onto StackQueue. There is an operation for count in StackQueue, but I won't be needing it for what I have planned for this queue, nor do any future planned uses.

If a need comes up we can add it then.

src/utils/tst/ThreadsafeBlockingQueue.cpp
jdelapla: That's exactly what the test right after this one does. multithreadTeardownTest

src/utils/include/com/amazonaws/kinesis/video/utils/Include.h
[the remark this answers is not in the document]
MushMal: The code will rot very fast if it's not agile to take changes like these. Higher level libraries have versioning for that purpose. Sure, it can cause a set of changes but all those changes are syntactical in nature. At the end of the day it's up to the owners of the codebase (which I am not) but I would recommend having a solid posture on how you keep the codebase agile even at the lowest layers.

src/utils/src/Semaphore.c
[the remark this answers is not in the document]
MushMal: My recommendation is to first think about your update story. Utility library was not versioned as it was going to be statically linked to higher-level libraries. It was later exposed publicly so it might be indeed be used by client applications so it needs to be backward compatible. In this case, an introduction of another API would be advisable - something like an ...Ex version which takes extra parameters in a form of a versioned struct (very heavy) or a better yet adding a version to the API - blahBlahV2 with it's own param set. All are pretty ugly but necessary for any public API

src/utils/src/ThreadsafeBlockingQueue.c
MushMal: This is more about resilience of the code. In the future the failure modes can be more diverse so accounting for it is a good idea.

src/utils/src/ThreadsafeBlockingQueue.c
MushMal: Oh I see, I misread it to be not equal to STATUS_SUCCESS

src/utils/src/ThreadsafeBlockingQueue.c
MushMal: OK

MushMal:
Most of my comments are lower priority after the last iteration

src/utils/src/Semaphore.c
```diff
@@ -88,7 +113,12 @@ STATUS semaphoreCreateInternal(UINT32 maxPermits, PSemaphore* ppSemaphore)
...
+    if (empty) {
+        ATOMIC_STORE(&pSemaphore->permitCount, 0);
```
disa6302, line 118: If the permitCount is set to 0, no resource would actually be able to acquire the semaphore per se right? 

This is the flow I am seeing when a thread tries to acquire the semaphore:
```
semaphoreAcquireInternal(...) {
    existingCount = ATOMIC_DECREMENT(&pSemaphore->permitCount); ---> existingCount = 0, permitCount = -1
    if ((INT32) existingCount <= 0) { ---> this evaluates to TRUE
    acquireFailed = TRUE; 
 CleanUp: 
     if (acquireFailed) { ---> TRUE
        ATOMIC_INCREMENT(&pSemaphore->permitCount); ---> permitCount = 0 since it was decremented to -1 previously
    }
```

Which means semaphore acquire would never succeed. Is that the intention?

src/utils/src/ThreadsafeBlockingQueue.c
disa6302: With setting permit count to 0, this will actually be an infinite loop unless pSafeQueue->terminate is set to TRUE while freeing the queue. Check comment above.

src/utils/src/Semaphore.c
jdelapla: It won't succeed until semaphoreRelease is called. It's mechanically the same as blocking/waiting for a signal.

src/utils/src/ThreadsafeBlockingQueue.c
jdelapla: Not quite. When the `existingCount <= 0` the code proceeds to line ~198, this line:

`CHK_STATUS(CVAR_WAIT(pSemaphore->semaphoreNotify, pSemaphore->semaphoreLock, timeout));`
Here the timeout is 0, so it will immediately timeout and the status won't be STATUS_SUCCESS, it will be some variation of STATUS_TIMEOUT.

disa6302:
Seems like some unit tests are still failing.

jdelapla:
Closing as threadpool needed changes to threadsafeblockingqueue, having them as one PR.
https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/199

PR #197, opened 2023-07-18
Fix ubsan failure

disa6302:
What was changed?

1. The size calculation of frame data is adapted to calculate and store size in a new variable instead of traversing the pointer if the AdaptedFrameData pointer is set to NULL.
2. Enabled stack trace for UBSAN CI job to make it easier to understand the source of failure.

Why was it changed?

We were doing arithmetic (pointer manipulation) when set to NULL, which is undefined behavior. It does seem that modern compilers seem to handle this fine without leading to a segfault, but it is better to ensure no arithmetic operation or pointer advancements are done on NULL pointers.

How was it changed?

Added a new local variable to calculate size of frame data if the pointer is NULL.

Testing:

Tested with the existing unit tests on which the UBSAN was failing.


hassanctech (commented):
This is really great!  Can you please add a test that fails with the current code but passes with this fix?

disa6302:
I did not add them because existing unit tests in itself fail and are pretty extensive passing in different arguments. The tests that test this API out are `AnnexBCpdNalAdapterTest`, `AnnexBNalAdapterTest` and `AudioCpdParserTest`. Let me know if there is a specific check you are looking for and I can add that in.

hassanctech (approved):
Nice!  

+1 for adding the gcc 4.4 back !!

PR #200, opened 2023-07-26
Log the content type when they don't match

daveisfera:
When running the sample with H.265, this warning is output, but it doesn't provide details of what the mismatch is, so this resolves that


hassanctech (approved):
lgtm

PR #201, opened 2023-07-27
Remove INLINE from functions that are declared but not defined becaus…

daveisfera:
Cleans up the build a bit so it's easier to see what's going on when things break


daveisfera:
I added this to see why some of the checks were failing but it appears this was actually the cause of why they were failing, so double win on this one

disa6302 (approved):

disa6302:
Thanks @daveisfera !

PR #215, opened 2023-09-11
Use const PCHAR so C++ compilers won't complain about passing string constants

daveisfera:
C++ compilers complain about passing string constants to PCHAR so this adds `const` to those so the use of the value is explicit and avoids those issues

### DIFF
--- a/src/common/include/com/amazonaws/kinesis/video/common/PlatformUtils.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/PlatformUtils.h
@@ -22,12 +22,12 @@ extern "C" {
 #endif
 
 // Log print function definition
-typedef VOID (*logPrintFunc)(UINT32, PCHAR, PCHAR, ...);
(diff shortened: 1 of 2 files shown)


disa6302 (commented):
Thank you for the PR! Could you direct this to develop so that I can merge it in?

daveisfera:
Done

disa6302:
Thanks @daveisfera . I am seeing your PR is failing clang format check. 
Can you clang format the code and push a new commit? 
You need to run this from the SDK root directory:

`./scripts/check-clang.sh` : This will give details on the affected files. 
`./scripts/clang-format.sh -f <file-name-with-path`: The file name would be the one that got returned from the previous command.

disa6302 (dismissed):

daveisfera:
I ran the check and none of the fixes I'm seeing are related to this PR

disa6302 (approved):

PR #220, opened 2023-09-19
Add stream state machine iterator

stefankiesz:
## Overview
Removes the possibility of segfault errors occurring from stack overflows caused by recursive `stepStateMachine()` calls made internally by the stream pointer functions passed as `executeStateFn` to the stream state machine. Implements the following changes:

- Iterating function `iterateStreamStateMachine()` added to the **StreamState** file
- Boolean member `keepIteratingStateMachine` added to the **KinesisVideoStream** struct to track when to keep calling `stepStateMachine()`
- All `stepStateMachine()` calls removed from `executeStateFn` functions
- `iterateStreamStateMachine()` makes `stepStateMachine()` calls as needed based on how was done in the `executeStateFn` functions
- All `stepStateMachine()` calls upon the stream state machine that are outside of `iterateStreamStateMachine()` replaced by `iterateStreamStateMachine()` calls

The functionality of the internal stepping of the state machine remains similar to the existing functionality, but the `executeStateFn` functions now set the `keepIteratingStateMachine` member to TRUE rather than calling `stepStateMachine()`.

### Stream State Behavior for Reference
- **1** state that _always_ internally calls `stepStateMachine()`: NEW
- **2** states that _sometimes_ internally call `stepStateMachine()`: READY, STOPPED
- **7** states that _never_ internally call `stepStateMachine()`: DESCRIBE, CREATE, TAG_STREAM, GET_ENDPOINT, GET_TOKEN, PUT_STREAM, STREAMING

## Testing
To test these changes' success in preventing the state transition segfaults, the state machine was forced into an endless loop by commenting out the following line in `fromNewStreamState()`:
```
state = STREAM_STATE_DESCRIBE;
```
The state machine would then endlessly transition from and to STREAM_STATE_NEW.

Without the iterator, this almost immediately caused a segfault, not even reaching 40,000 recursive calls:
Screenshot 2023-09-19 at 2 59 36 PM

Using the iterator, the sate machine continues to transition, reaching over 4 million iterations in ~1min:
Screenshot 2023-09-19 at 2 46 56 PM

<br>
<br>

### DIFF
--- a/src/client/src/StreamState.c
+++ b/src/client/src/StreamState.c
@@ -50,7 +50,7 @@ STATUS iterateStreamStateMachine(PKinesisVideoStream pKinesisVideoStream)
     PStateMachine pStateMachine = NULL;
 
     CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
-    
+
     pStateMachine = pKinesisVideoStream->base.pStateMachine;
     CHK(pStateMachine != NULL, STATUS_NULL_ARG);
 
--- a/src/client/src/StreamState.c
+++ b/src/client/src/StreamState.c
@@ -56,7 +56,7 @@ STATUS iterateStreamStateMachine(PKinesisVideoStream pKinesisVideoStream)
     do {
         pKinesisVideoStream->keepIteratingStateMachine = FALSE;
         CHK_STATUS(stepStateMachine(pStateMachine));
-        CHK(pKinesisVideoStream != NULL && pStateMachine != NULL, STATUS_NULL_ARG);
+        CHK(pKinesisVideoStream != NULL, STATUS_NULL_ARG);
     } while (pKinesisVideoStream->keepIteratingStateMachine);
 
 CleanUp:
--- a/src/client/src/StreamState.c
+++ b/src/client/src/StreamState.c
@@ -50,9 +50,9 @@ STATUS iterateStreamStateMachine(PKinesisVideoStream pKinesisVideoStream)
     PStateMachine pStateMachine = pKinesisVideoStream->base.pStateMachine;
 
     do {
-        pKinesisVideoStream->keepIterating = FALSE;
+        pKinesisVideoStream->keepIteratingStateMachine = FALSE;
         CHK_STATUS(stepStateMachine(pStateMachine));
-    } while (pKinesisVideoStream->keepIterating);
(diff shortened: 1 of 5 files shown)


src/state/src/State.c
```diff
@@ -161,7 +161,7 @@ STATUS stepStateMachine(PStateMachine pStateMachine)
 
     // Check if we are changing the state
     if (pState->state != pStateMachineImpl->context.pCurrentState->state) {
-        // Since we're transitioning to a different state from this state, reset the local state retry count to0
+        // Since we're transitioning to a different state from this state, reset the local state retry count to 0
```
sirknightj, line 164: Nice

src/client/src/StreamState.c
```diff
@@ -39,6 +39,27 @@ StateMachineState STREAM_STATE_MACHINE_STATES[] = {
...
 
+///////////////////////////////////////////////////////////////////////////
+// State machine transition iterator
+///////////////////////////////////////////////////////////////////////////
+
+STATUS iterateStreamStateMachine(PKinesisVideoStream pKinesisVideoStream)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PStateMachine pStateMachine = pKinesisVideoStream->base.pStateMachine;
```
jdelapla, line 50: Add null checks for all pointers used in this function.

jdelapla (changes requested):

src/client/src/StreamState.c
stefankiesz: Added!

jdelapla (approved):

hassanctech (changes requested):
Before we can merge this PR, I would like to request can you please create a PR in producer C against master / latest release, update the PIC git hash to the one from this PR, so the CI can run in Producer C and we make sure they are all passing with this version of PIC.  The Producer C CI tests are a much better indicator of whether all the state transitions are working properly than the tests inside PIC itself.  It has happened in the past we have made changes at this layer and not known we messed stuff up until we ran producer c which has a lot more coverage.

hassanctech (dismissed):

jdelapla (approved):

jdelapla (changes requested):

jdelapla (approved):

PR #229, opened 2023-12-08
Update CommonDefs.h

hassanctech:
*Issue #, if available:* N/A

*Description of changes:*
The way the PIC layer decides whether a platform is 64 or 32 bit is by examining some pre-defined compiler macros.  For 64 bad sadly we had a big miss that arm64 was entirely omitted, i.e. `__aarch64__` as specified here:  https://sourceforge.net/p/predef/wiki/Architectures/

This is pretty bad because this maps our SIZE_T macro to a 32 bit value instead of 64 bit as the native `size_t` on the platform.  This PR resolves that issue.


disa6302 (approved):
Awesome!

PR #230, opened 2023-12-08
Fix 32 bit crash issue

hassanctech:
*Issue #, if available:*

*Description of changes:*
In `threadpoolFree` there is a call made to `stackQueueIteratorGetItem` which pass the address of a pointer, but the API expected and returns a UINT64 in this field, this will cause a crash on 32 bit platforms.  For reference similar issue's have been fixed and proper usage can be found elsewhere in PIC:  https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/57637ea593f4b43c509413a44d993ed08d7f2616/src/client/src/Stream.c#L2059-L2062

I've modified this to follow the same pattern.  The CHK for NULL was missing as well if it's null we will seg fault when we try to access `item->dataMutex`.

### DIFF
--- a/src/utils/src/Threadpool.c
+++ b/src/utils/src/Threadpool.c
@@ -291,7 +291,8 @@ STATUS threadpoolFree(PThreadpool pThreadpool)
     STATUS retStatus = STATUS_SUCCESS;
     StackQueueIterator iterator;
     PThreadData item = NULL;
-    BOOL finished = FALSE, taskQueueEmpty = FALSE;
+    UINT64 data;
+    BOOL finished = FALSE, taskQueueEmpty = FALSE, listMutedLocked = FALSE;
     CHK(pThreadpool != NULL, STATUS_NULL_ARG);
 
     // Threads are not forced to finish their tasks. If the user has assigned
@@ -311,6 +312,7 @@ STATUS threadpoolFree(PThreadpool pThreadpool)
     while (!finished) {
         // lock list mutex
         MUTEX_LOCK(pThreadpool->listMutex);
+        listMutedLocked = TRUE;
 
         do {
             // iterate on list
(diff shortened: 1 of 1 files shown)


src/utils/src/Threadpool.c
```diff
@@ -319,15 +321,18 @@ STATUS threadpoolFree(PThreadpool pThreadpool)
                 finished = TRUE;
                 break;
             }
-            retStatus = stackQueueIteratorGetItem(iterator, &item);
+
+            CHK_STATUS(stackQueueIteratorGetItem(iterator, &data));
+            item = (PThreadData) data;
+            CHK(item != NULL, STATUS_INTERNAL_ERROR);
```
jdelapla, line 327: If the iterator is still valid, this means it may not be the last item in the thread list. We would need to handle all other items before exiting this loop, or the corresponding threads will not terminate gracefully.

jdelapla (changes requested):

src/utils/src/Threadpool.c
hassanctech: The paradigm currently employed is the following sequence of calls in a loop:
```
stackQueueGetIterator(pThreadpool->threadList, &iterator);
stackQueueIteratorGetItem(iterator, &data)
stackQueueRemoveItem(pThreadpool->threadList, data)
```

In order for the iterator to actually iterate we need to remove the item.  However the removal happens under the `item->dataMutex` lock... but this is the very thing that will crash if `item` is `NULL`.  Are you saying that it's actually possible for there to be `PThreadData` on the stack/queue with value of `NULL` and they need to be discarded and we need to keep iterating? 

If I understood correctly then I suppose that is possible by checking if item is NULL instead of returning a status null error we just remove the NULL item and keep going, without attempting to acquire the mutex or setting the terminate flag (since there is no actual item object available).  

Ideally it should not be possible to have `NULL` items on this stackQueue, but I'm not sure if that was part of your design for the threadpool?

hassanctech: Looking at this block it looks like it is not possible to ever have a NULL (0) item as a data in this queue:

```
    data->dataMutex = MUTEX_CREATE(FALSE);
    data->pThreadpool = pThreadpool;
    ATOMIC_STORE_BOOL(&data->terminate, FALSE);

    CHK_STATUS(stackQueueEnqueue(pThreadpool->threadList, (UINT64) data));
```

But if we did have `NULL` somehow that would imply some of corruption and in that case returning an internal failure seems appropriate.  Alternatively we could "remove" the `NULL` items and keep going as suggested earlier?    

The only way the `stackQueueIteratorGetItem(iterator, &data)` can error out is if the iterator itself is `NULL` in which case it is not valid and terminating is the correct thing to do.  Or if `&data` is `NULL` which is impossible.

jdelapla: It shouldn't be possible for item to be NULL unless something has gone wrong with the stackqueue. However since we're checking for it we should fully handle the use case.

If for whatever reason the item is null, but the iterator is still valid from the earlier check then we need to remove the item OR call `stackQueueIteratorNext()` until we've fully gone through the list.

It's not important that the item mutex be locked when removing it from the stack queue if the item itself is NULL.

hassanctech: Sure.  Previously before adding the check it would have crashed, now I've modified it to remove the NULL item and keep iterating. I added a WARN level log message if it is actually ever NULL since as you said that should not be possible.

PR #232, opened 2023-12-14
address compiler warnings, cleanup ci

hassanctech:
*Issue #, if available:* n/a

*Description of changes:*
1. Remove unnecessary double compiles in CI
2. Move test ci run into a its own step
3. Address many compiler warnings (zero warnings on mac and linux builds now with gcc and clang)
4. Modify build of test project to no longer re-build all of PIC again in order to set custom compile options for the test, since using `add_subdirectory` propagated the flag from PIC itself down to the test project so it will always match.
5. Refactor test project to be in it's own high level tst directory.  In addition we build and run tests for PIC with two configurations with and without the flags:  ALIGNED_MEMORY_MODEL and FIXUP_ANNEX_B_TRAILING_NALU_ZERO.  Previously the test project enabled both flags and we did not test the path with both flags disabled (i.e. the default configuration of this library).  In doing so many tests needed to be if/def'd behind these flags because they other wise crash or report the unaligned access by the usbsan run.  In the run with the aforementioned flags set, exactly the same suite of tests as before run so there has been no reduction in test coverage, rather what we have now is a superset of what we had previously.
6. Fix threadpool issue where in `threadpoolInternalInactiveThreadCount` we may subtract two unsigned numbers (A - B) where A < B and return the result.  Now we return 0 in such cases.


src/utils/src/Allocators.c
```diff
@@ -53,7 +53,7 @@ VOID dumpMemoryHex(PVOID pMem, UINT32 size)
     PCHAR pCur = buf;
     PBYTE pByte = (PBYTE) pMem;
     for (UINT32 i = 0; i < size; i++) {
-        SPRINTF(pCur, "%02x ", *pByte++);
+        SNPRINTF(pCur, 2, "%02x ", *pByte++);
```
disa6302, line 56: Isnt this supposed to be 3 to include the NULL character?
https://cplusplus.com/reference/cstdio/snprintf/

tst/client/AcksFunctionalityTest.cpp
```diff
@@ -21,7 +21,7 @@ class AcksFunctionalityTest : public ClientTestBase,
         mStreamInfo.streamCaps.replayDuration = (UINT64) replayDuration;
     }
 };
-
+#ifdef ALIGNED_MEMORY_MODEL
```
disa6302, line 24: Why do these tests require `ALIGNED_MEMORY_MODEL`?

tst/client/CallbacksAndPressuresFunctionalityTest.cpp
```diff
@@ -22,7 +22,7 @@ class CallbacksAndPressuresFunctionalityTest : public ClientTestBase,
         mStreamInfo.streamCaps.replayDuration = (UINT64) replayDuration;
     }
 };
-
+#ifdef ALIGNED_MEMORY_MODEL
```
disa6302, line 25: Same here.

disa6302 (commented):
I understand that all our CI had ALIGNED_MEMORY_MODEL enabled by default. It does seem odd that there are a lot of tests that have aligned memory model required, considering that it is an option that needs to be explicitly enabled by customers. What happens without this option on the unit tests? Do they crash?

disa6302 (approved):

PR #233, opened 2023-12-28
State machine name for logs

disa6302:
*Issue #, if available:*

*Description of changes:*

*What was changed?*

- Added a new API to allow setting tags for state machine logs.

*Why was it changed?*

Our SDKs have multiple state machines running at once and it only logs the state in hex, which makes it confusing to understand which state machine transitions are being tracked while debugging. This PR introduces addition of tags when logging state machine transitions in VERBOSE mode to make it easy to understand which state machine transition the logs belong to.

*How was it changed?*
Introduced a new API `createStateMachineWithTag` to pass in a tag. This is mandatory. if not provided the API call would fail.

Earlier:

```
2023-12-28 18:42:08.694 VERBOSE stepStateMachine(): State Machine - Current state: 0x0000000000000001, Next state: 0x0000000000000002, Current local state retry count [0], Max local state retry count [5], State transition wait time [0] ms
```

After this change:

```
2023-12-28 18:40:18.229 VERBOSE   stepStateMachine(): [STREAM] State Machine - Current state: 0x0000000000000010, Next state: 0x0000000000000040, Current local state retry count [0], Max local state retry count [5], State transition wait time [0] ms
```

*Testing`*
Included API unit tests to confirm positive and negative test cases.

### DIFF
--- a/src/state/src/State.c
+++ b/src/state/src/State.c
@@ -302,7 +302,8 @@ CleanUp:
     return retStatus;
 }
 
-STATUS setStateMachineTag(PStateMachine pStateMachine, PCHAR stateTag) {
+STATUS setStateMachineTag(PStateMachine pStateMachine, PCHAR stateTag)
+{
     STATUS retStatus = STATUS_SUCCESS;
     PStateMachineImpl pStateMachineImpl = (PStateMachineImpl) pStateMachine;
     CHK_WARN(pStateMachineImpl != NULL, STATUS_NULL_ARG, "State machine object not created. Cannot set tag");
--- a/src/state/src/State.c
+++ b/src/state/src/State.c
@@ -171,7 +171,7 @@ STATUS stepStateMachine(PStateMachine pStateMachine)
         pStateMachineImpl->context.localStateRetryCount++;
     }
 
-    DLOGD("[%s] State Machine - Current state: 0x%016" PRIx64 ", Next state: 0x%016" PRIx64 ", "
+    DLOGV("[%s] State Machine - Current state: 0x%016" PRIx64 ", Next state: 0x%016" PRIx64 ", "
           "Current local state retry count [%u], Max local state retry count [%u], State transition wait time [%u] ms",
           pStateMachineImpl->stateTag, pStateMachineImpl->context.pCurrentState->state, nextState, pStateMachineImpl->context.localStateRetryCount,
           pState->maxLocalStateRetryCount, errorStateTransitionWaitTime / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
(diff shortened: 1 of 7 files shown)


hassanctech:
This is super useful, thanks for doing this @disa6302 !  I have a concern though about truncation if the user specifies a tag longer than the max allowed length.  It feels wrong to return a status success when we failed to use the tag specified by the user.  I think in this case we should return a failure status and not set a tag at all.  If no tag is set we should not have the extra `[]`.  It would be nice to make the tag required but would break backwards compat...

disa6302:
> This is super useful, thanks for doing this @disa6302 ! I have a concern though about truncation if the user specifies a tag longer than the max allowed length. It feels wrong to return a status success when we failed to use the tag specified by the user. I think in this case we should return a failure status and not set a tag at all. If no tag is set we should not have the extra `[]`. It would be nice to make the tag required but would break backwards compat...

I thought about this. I really wanted to add it as part of `createStateMachine` and not have a separate API. I am ok with adding the return code to fail on this, but given its a tag to uniquely identify a state machine log line, I thought it is ok to truncate (something is better than nothing ideology) instead of setting the tag again in case of failure.

src/state/src/State.c
```diff
@@ -54,6 +54,58 @@ STATUS createStateMachine(PStateMachineState pStates, UINT32 stateCount, UINT64
     return retStatus;
 }
 
+/**
+ * Creates a new state machine
+ */
```
sirknightj, line 59: Let's document what is the behavior when the tag is invalid or too long

src/state/src/State.c
```diff
@@ -54,6 +54,55 @@ STATUS createStateMachine(PStateMachineState pStates, UINT32 stateCount, UINT64
...
+                                  UINT64 getCurrentTimeFuncCustomData, PCHAR pStateMachineName, PStateMachine* ppStateMachine)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PStateMachineImpl pStateMachine = NULL;
+    UINT32 allocationSize = 0;
+
+    CHK(pStates != NULL && ppStateMachine != NULL && getCurrentTimeFunc != NULL && pStateMachineName != NULL, STATUS_NULL_ARG);
+    CHK(stateCount > 0, STATUS_INVALID_ARG);
+    CHK(STRLEN(pStateMachineName) <= MAX_STATE_MACHINE_NAME_LENGTH, STATUS_STATE_MACHINE_NAME_LEN);
```
sirknightj, line 67: May want to use STRNLEN for additional safety, like this:
```
CHK(STRNLEN(formatStr, MAX_TIMESTAMP_FORMAT_STR_LEN + 1) <= MAX_TIMESTAMP_FORMAT_STR_LEN, STATUS_MAX_TIMESTAMP_FORMAT_STR_LEN_EXCEEDED);
```

sirknightj (approved):

src/state/src/State.c
hassanctech: Yeah we absolutely need this check since this is client supplied and we've only checked it isn't null... it could be non null terminated at all and result in a memory access violation with STRLEN.

src/state/src/State.c
```diff
@@ -54,6 +54,55 @@ STATUS createStateMachine(PStateMachineState pStates, UINT32 stateCount, UINT64
     return retStatus;
 }
 
+STATUS createStateMachineWithName(PStateMachineState pStates, UINT32 stateCount, UINT64 customData, GetCurrentTimeFunc getCurrentTimeFunc,
```
hassanctech, line 57: In other implementation we have like this, basically the *only* difference between this and `createStateMachineWithName` and `createStateMachine` is that we are setting the name in the former.  I think it would be much cleaner here to call `createStateMachine` inside this function instead of re-doing (copy/paste) all that work here, and then just set the name as well, perhaps with the returned state machine object, or however is most convenient.

hassanctech (approved):
nice :)

PR #234, opened 2024-01-03
Add LOG_LEVEL_SILENT explicitly to avoid PROFILE logs from getting displayed when setting log level to SILENT

disa6302:
*Issue #, if available:*

*Description of changes:*

A new log level was introduced in previous version where the value of the level is set to `LOG_LEVEL_PROFILE` (8). LOG_LEVEL_SILENT is set to 7. The SDK has a filtering logic where we log messages with log level greater than the set logLevel. This means, with `LOG_LEVEL_SILENT`, no log messages should be displayed. However, with the new added `LOG_LEVEL_PROFILE`, this breaks when setting log level to `LOG_LEVEL_SILENT` since PROFILE logs would get displayed. This PR fixes the issue.

Testing:

- Tested with WebRTC C sample (which uses PROFILE log level) by setting LOG_LEVEL_SILENT to confirm no logs get displayed if log level is set to 7. Also tested setting every other level.


stefankiesz (approved):
LGTM!

sirknightj (approved):

PR #238, opened 2024-01-17
Threadpool tsan

jdelapla:
*Issue #, if available:*
N/A

*Description of changes:*
ThreadData->mutex now wraps all interactions with any threadpool objects, this includes the queue.
As a consequence of having the data mutex locked while waiting on the queue, the threadpool couldn't teardown if an actor was waiting on the queue, so the teardown now pushes wait tasks to the queue in the event of that use case.

Additionally all utils tests were changed so that if a memory leak is detected, it will infinitely wait for the memory leak to end -- resulting in a GHA timeout or a loose actor cleaning up the memory (common race condition specific to tests only).


disa6302:
Can we open a PR in WebRTC to confirm the CI there passes too?

hassanctech (approved):
lgtm

PR #240, opened 2024-01-25
safe gmtime, guard with mutex since not thread safe on non windows pl…

hassanctech:
*Issue #, if available:* Address TSAN failures in Producer C

*Description of changes:*
C function `gmtime` is not threadsafe on non-windows platforms.  Create a new `GMTIME_SAFE` macro which guards the gmtime call with a global mutex.

The if/def for this includes CYGWIN because I'm using pthread which our Thread platform dependent layer does not use for cygwin.

### DIFF
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -997,7 +997,7 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 #define STRFTIME    strftime
 #define GMTIME      gmtime
 
-#if defined _WIN32 || defined _WIN64
+#if defined _WIN32 || defined _WIN64 || defined __CYGWIN__
 #define GMTIME_SAFE GMTIME
 #else
 #define GMTIME_SAFE globalGetTmTime
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -694,7 +694,15 @@ typedef struct tm* (*getTmTime)(const time_t*);
 #define TIME_DIFF_UNIX_WINDOWS_TIME 116444736000000000ULL
 
 PUBLIC_API UINT64 defaultGetTime();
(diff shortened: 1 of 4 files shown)


src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -686,18 +686,22 @@ extern getTName globalGetThreadName;
...
 
+typedef struct tm* (*getTmTime)(const time_t*);
+
 //
 // Default time library functions
 //
 #define TIME_DIFF_UNIX_WINDOWS_TIME 116444736000000000ULL
 
 PUBLIC_API UINT64 defaultGetTime();
+PUBLIC_API struct tm* defaultGetTmTime(const time_t*);
```
disa6302, line 697: Since it is public definition, I think we should name it `defaultGetThreadSafeTime` or something similar to be explicit. Also, can we add an API explanation?

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -993,6 +997,12 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 #define STRFTIME    strftime
 #define GMTIME      gmtime
 
+#if defined _WIN32 || defined _WIN64 || defined __CYGWIN__
+#define GMTIME_SAFE GMTIME
+#else
+#define GMTIME_SAFE globalGetTmTime
```
disa6302, line 1003: Nit: `GMTIME_THREAD_SAFE`

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
hassanctech: Good point, I agree.  How about `defaultGetThreadSafeTmTime` I want the tm in the name just so it's different from GetTime -- it doesn't return `UINT64` time instead it returns the `tm` struct.

disa6302: Sure! That works!

disa6302 (approved):

PR #242, opened 2024-02-03
unlock stream mutex before acquiring the fom mutex to avoid deadlock …

hassanctech:
*Issue #, if available:*

the PutFrame path acquires locks in the following order:

`putFrameLock -> frameOrderCoordinatorLock -> StreamLock -> ClientLock`

The free path acquires lock in the following order:

`streamListLock -> StreamLock -> FrameOrderCoordinatorLock`

Hence here is a possibility of deadlock

*Description of changes:*

Unlock the stream lock *before* acquiring the frame order coordinator lock in the freeStream path in order to maintain a consistent lock acquisition order and avoid a deadlock.

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -363,8 +363,15 @@ STATUS freeStream(PKinesisVideoStream pKinesisVideoStream)
     freeMetadataTracker(&pKinesisVideoStream->eosTracker);
     freeMetadataTracker(&pKinesisVideoStream->metadataTracker);
 
+    // unlock the stream freeFrameOrderCoordinator acquires the FrameOrderCoordinator mutex, we cannot acquire this
+    // mutex while holding the stream mutex or else we will deadlock with putFrame which first acquires
+    // FrameOrderCoordinator mutex and then the stream mutex
+    pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+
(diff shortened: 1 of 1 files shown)


src/client/src/Stream.c
```diff
@@ -363,9 +363,16 @@ STATUS freeStream(PKinesisVideoStream pKinesisVideoStream)
...
+    // unlock the stream freeFrameOrderCoordinator acquires the FrameOrderCoordinator mutex, we cannot acquire this
+    // mutex while holding the stream mutex or else we will deadlock with putFrame which first acquires
+    // FrameOrderCoordinator mutex and then the stream mutex
+    pKinesisVideoClient->clientCallbacks.unlockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+
     // Free FrameOrderCoordinator
     freeFrameOrderCoordinator(pKinesisVideoStream, &pKinesisVideoStream->pFrameOrderCoordinator);
 
+    pKinesisVideoClient->clientCallbacks.lockMutexFn(pKinesisVideoClient->clientCallbacks.customData, pKinesisVideoStream->base.lock);
+
```
niyatim23, line 375: We acquire the lock again in the next line. Is that because previously some function on the path unlocked the mutex already?

hassanctech: The assumption in this function is that we have the lock LN 406-407 we unlock and free this lock.

disa6302 (approved):

PR #243, opened 2024-02-07
Add API for specifying thread stack size

jdelapla:
Issue #, if available:

*What was changed?*
Added API to support specifying stack size of threads at time of creation.
Added an ability to specify default stack sizes at compile time.

*Why was it changed?*
Libraries that use this asset as a dependency needed to be able to customize their thread stack sizes

*How was it changed?*
New API added, check for default compile time variable.

Old tests passing, new tests added. Additional changes made in upstream libraries using this change that also have tests passing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,7 @@ function(enableSanitizer SANITIZER)
 endfunction()
 
 if(KVS_DEFAULT_STACK_SIZE)
-  message(STATUS "@@@@@@@@@@@@@@@@@@@@Building with default stack size")
+  message(STATUS "Building with default stack size")
   add_compile_definitions(KVS_DEFAULT_STACK_SIZE=${KVS_DEFAULT_STACK_SIZE})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,10 @@ function(enableSanitizer SANITIZER)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=${SANITIZER}" PARENT_SCOPE)
 endfunction()
 
+if(KVS_DEFAULT_STACK_SIZE)
+  add_compile_definitions(KVS_DEFAULT_STACK_SIZE=${KVS_DEFAULT_STACK_SIZE})
+endif()
+
 if(ADDRESS_SANITIZER)
   enableSanitizer("address")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,7 @@ function(enableSanitizer SANITIZER)
 endfunction()
 
 if(KVS_DEFAULT_STACK_SIZE)
+  message(STATUS "@@@@@@@@@@@@@@@@@@@@Building with default stack size")
   add_compile_definitions(KVS_DEFAULT_STACK_SIZE=${KVS_DEFAULT_STACK_SIZE})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ endfunction()
 
 if(KVS_DEFAULT_STACK_SIZE)
   message(STATUS "Building with default stack size")
-  add_compile_definitions(KVS_DEFAULT_STACK_SIZE=${KVS_DEFAULT_STACK_SIZE})
+  add_compile_definitions(KVS_DEFAULT_STACK_SIZE_BYTES=${KVS_DEFAULT_STACK_SIZE})
 endif()
 
 if(ADDRESS_SANITIZER)
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ You can pass the following options to `cmake ..`
 * `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
 * `-DBUILD_DEBUG_HEAP` Build debug heap with guard bands and validation. This is ONLY intended for low-level debugging purposes. Default is OFF
 * `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
-* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16kb
+* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16 Kib
 
 ### Build
 To build the library run make in the build directory you executed CMake.
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ You can pass the following options to `cmake ..`
 * `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
 * `-DBUILD_DEBUG_HEAP` Build debug heap with guard bands and validation. This is ONLY intended for low-level debugging purposes. Default is OFF
 * `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
+* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16kb
 
 ### Build
 To build the library run make in the build directory you executed CMake.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ You can pass the following options to `cmake ..`
 * `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
 * `-DBUILD_DEBUG_HEAP` Build debug heap with guard bands and validation. This is ONLY intended for low-level debugging purposes. Default is OFF
 * `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
-* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16 Kib
+* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16 Kib.
 
 ### Build
 To build the library run make in the build directory you executed CMake.
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -721,6 +721,7 @@ typedef VOID (*unlockMutex)(MUTEX);
 typedef BOOL (*tryLockMutex)(MUTEX);
 typedef VOID (*freeMutex)(MUTEX);
 typedef STATUS (*createThread)(PTID, startRoutine, PVOID);
+typedef STATUS (*createThreadWithParams)(PTID, startRoutine, SIZE_T, PVOID);
 typedef STATUS (*joinThread)(TID, PVOID*);
 typedef VOID (*threadSleep)(UINT64);
 typedef VOID (*threadSleepUntil)(UINT64);
@@ -763,6 +764,7 @@ extern unlockMutex globalUnlockMutex;
 extern tryLockMutex globalTryLockMutex;
 extern freeMutex globalFreeMutex;
 extern createThread globalCreateThread;
+extern createThreadWithParams globalCreateThreadWithParams;
 extern joinThread globalJoinThread;
 extern threadSleep globalThreadSleep;
 extern threadSleepUntil globalThreadSleepUntil;
@@ -1032,12 +1034,13 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 //
 // Thread functionality
 //
-#define THREAD_CREATE      globalCreateThread
-#define THREAD_JOIN        globalJoinThread
-#define THREAD_SLEEP       globalThreadSleep
-#define THREAD_SLEEP_UNTIL globalThreadSleepUntil
-#define THREAD_CANCEL      globalCancelThread
-#define THREAD_DETACH      globalDetachThread
+#define THREAD_CREATE             globalCreateThread
+#define THREAD_CREATE_WITH_PARAMS globalCreateThreadWithParams
+#define THREAD_JOIN               globalJoinThread
+#define THREAD_SLEEP              globalThreadSleep
+#define THREAD_SLEEP_UNTIL        globalThreadSleepUntil
+#define THREAD_CANCEL             globalCancelThread
+#define THREAD_DETACH             globalDetachThread
 
 //
 // Static initializers
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -1034,7 +1034,7 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 //
 // Thread functionality
 //
-//Wrappers around OS specific utilities for threads. Takes arguments as given.
+// Wrappers around OS specific utilities for threads. Takes arguments as given.
 #define THREAD_CREATE             globalCreateThread
 #define THREAD_CREATE_WITH_PARAMS globalCreateThreadWithParams
 #define THREAD_JOIN               globalJoinThread
--- a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
+++ b/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
@@ -1034,6 +1034,7 @@ extern PUBLIC_API atomicXor globalAtomicXor;
 //
 // Thread functionality
 //
+//Wrappers around OS specific utilities for threads. Takes arguments as given.
 #define THREAD_CREATE             globalCreateThread
 #define THREAD_CREATE_WITH_PARAMS globalCreateThreadWithParams
 #define THREAD_JOIN               globalJoinThread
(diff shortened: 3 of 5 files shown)


src/utils/src/Thread.c
```diff
@@ -150,16 +177,71 @@ PUBLIC_API TID defaultGetThreadId()
...
     CHK(pThreadId != NULL, STATUS_NULL_ARG);
 
-#ifdef CONSTRAINED_DEVICE
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    pthread_attr_t attr;
+    pAttr = &attr;
+    result = pthread_attr_init(pAttr);
+    CHK_ERR(result == 0, STATUS_THREAD_ATTR_INIT_FAILED, "pthread_attr_init failed with %d", result);
+    stackSize = KVS_DEFAULT_STACK_SIZE;
+    result = pthread_attr_setstacksize(&attr, stackSize);
```
hassanctech, line 242: No need for extra local variable:

```
result = pthread_attr_setstacksize(&attr, (SIZE_T) KVS_DEFAULT_STACK_SIZE);
```

src/utils/src/Thread.c
```diff
@@ -150,16 +177,71 @@ PUBLIC_API TID defaultGetThreadId()
...
+        result = pthread_attr_destroy(pAttr);
+        if (result != 0) {
+            DLOGW("pthread_attr_destroy failed with %u", result);
+        }
+    }
+
+    return retStatus;
+}
+
 PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
```
hassanctech, line 226: Would be ideal if this can call `defaultCreateThreadWithParams` and just specify the thread size it needs, so we don't have the rest of the logic in two places.

src/utils/src/Thread.c
disa6302: I think we should have a min check here and proactively set the stack size to min if the value is below 16KB. I see we have a check in our WebRTC samples, but instead of returning a fail status, might as well set to a default minimum. What do you think? 

If we do not want to default, it might be useful to have specific status codes for specific result value returned by `pthread_attr_setstacksize`.

src/utils/src/Thread.c
```diff
@@ -51,7 +51,7 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
     pWrapper->storedArgs = args;
     pWrapper->storedStartRoutine = start;
 
-    threadHandle = CreateThread(NULL, 0, startWrapperRoutine, pWrapper, 0, NULL);
+    threadHandle = CreateThread(NULL, stackSize, startWrapperRoutine, pWrapper, 0, NULL);
```
disa6302, line 54: Do we know what the default min stack size for Windows is? Can we have a unit test for all Linux and Ubuntu to test out the negative return codes too (specifically looking for setting a really low value to confirm the failure status code.

src/utils/src/Thread.c
```diff
@@ -198,6 +213,23 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
     return retStatus;
 }
 
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, KVS_DEFAULT_STACK_SIZE, args));
```
hassanctech, line 221: Can we explicit cast `SIZE_T` otherwise I'm pretty sure we'll have compiler warnings.  For example:
```
CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, (SIZE_T) KVS_DEFAULT_STACK_SIZE, args));
```

src/utils/src/Thread.c
```diff
@@ -64,6 +64,21 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
...
 }
 
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, (SIZE_T)KVS_DEFAULT_STACK_SIZE, args);
+#else
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, 0, args);
```
hassanctech, line 74: Missing closing `)`

CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ function(enableSanitizer SANITIZER)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=${SANITIZER}" PARENT_SCOPE)
 endfunction()
 
+if(KVS_DEFAULT_STACK_SIZE)
```
disa6302, line 61: Can we add this to the top of the cmake list file as a valid option: 

`option(KVS_DEFAULT_STACK_SIZE "Setting thread default stack size in bytes" OFF)`

CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ function(enableSanitizer SANITIZER)
...
+  message(STATUS "Building with default stack size")
```
disa6302, line 62: What happens here if the value is empty? We should have validation here. I see the validation in webrtc, but we should add it here too. 

Also, a nit: can we rename `KVS_DEFAULT_STACK_SIZE` to `KVS_DEFAULT_STACK_SIZE_BYTES` or something similar to be explicit about the unit.

hassanctech: +1

src/utils/src/Thread.c
jdelapla: I tried looking for that but I couldn't find a source

CMakeLists.txt
jdelapla: That would give it a default value of OFF woudln't it?

src/utils/src/Thread.c
jdelapla: There doesn't appear to be a minimum stack size for windows that I can find in any documentation

src/utils/src/Thread.c
jdelapla: done

src/utils/src/Thread.c
jdelapla: done

src/utils/src/Thread.c
jdelapla: done

src/utils/src/Thread.c
jdelapla: done

src/utils/src/Thread.c
jdelapla: done

CMakeLists.txt
jdelapla: If the value is empty it's undefined and goes to 'else'

jdelapla: renamed compile time variable to have _BYTES for readability.

src/utils/src/Thread.c
jdelapla: the failure status code of linux is 22, invalid parameter. For Windows it will just crash from stack overflow if you provide an incredibly low value.

src/utils/src/Thread.c
```diff
@@ -198,6 +213,23 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
     return retStatus;
 }
 
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, KVS_DEFAULT_STACK_SIZE, args));
+#elif defined(CONSTRAINED_DEVICE)
```
stefankiesz, line 227: Should we document/comment somewhere to point out that a specified default stack size will supersede the constrained device size?

src/utils/src/Thread.c
```diff
@@ -64,6 +64,21 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
     return retStatus;
 }
 
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, KVS_DEFAULT_STACK_SIZE, args);
```
stefankiesz: Missing closing parentheses on these?

src/utils/src/Thread.c
```diff
@@ -198,6 +213,23 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
...
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, KVS_DEFAULT_STACK_SIZE, args));
+#elif defined(CONSTRAINED_DEVICE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, THREAD_STACK_SIZE_ON_CONSTRAINED_DEVICE, args));
+#else
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, 0, args);
```
stefankiesz: Missing closing parentheses? Not sure how this wouldn't have failed to build though

stefankiesz (changes requested):

src/utils/src/Thread.c
jdelapla: already fixed

src/utils/src/Thread.c
jdelapla: already fixed

src/utils/src/Thread.c
jdelapla: Added to producer-c where CONSTRAINED_DEVICE is listed.

README.md
```diff
@@ -55,6 +55,7 @@ You can pass the following options to `cmake ..`
 * `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
 * `-DBUILD_DEBUG_HEAP` Build debug heap with guard bands and validation. This is ONLY intended for low-level debugging purposes. Default is OFF
 * `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
+* `-DKVS_DEFAULT_STACK_SIZE` -- Value in bytes for default stack size of threads. Pthread minimum for Unix applications is 16kb
```
sirknightj, line 58: kib since multiple of 1024

tst/utils/Thread.cpp
```diff
@@ -117,3 +117,38 @@ TEST_F(ThreadFunctionalityTest, ThreadCreateAndCancel)
 
     MUTEX_FREE(gThreadMutex);
 }
+
+TEST_F(ThreadFunctionalityTest, ThreadCreateAndReleaseSimpleCheckWithStack)
+{
+    UINT64 index;
+    TID threads[TEST_THREAD_COUNT];
+    gThreadMutex = MUTEX_CREATE(FALSE);
+    SIZE_T threadStack = 16 * 1024;
```
sirknightj, line 126: For sanity can we choose something bigger than the minimum?

tst/utils/Thread.cpp
```diff
@@ -117,3 +117,38 @@ TEST_F(ThreadFunctionalityTest, ThreadCreateAndCancel)
...
+        EXPECT_EQ(STATUS_SUCCESS, THREAD_JOIN(threads[index], NULL));
+    }
+
+    MUTEX_LOCK(gThreadMutex);
+    EXPECT_EQ(0, gThreadCount);
+    MUTEX_UNLOCK(gThreadMutex);
+
+    for (index = 0; index < TEST_THREAD_COUNT; index++) {
+        EXPECT_TRUE(st[index].threadVisited) << "Thread didn't visit index " << index;
+        EXPECT_TRUE(st[index].threadCleared) << "Thread didn't clear index " << index;
```
sirknightj, line 150: May want to add a TODO to add a check that the size is correct

tst/utils/Thread.cpp
```diff
@@ -117,3 +117,38 @@ TEST_F(ThreadFunctionalityTest, ThreadCreateAndCancel)
...
+    }
+
+    MUTEX_FREE(gThreadMutex);
+}
```
sirknightj, line 154: Can we also add a negative case where it didn't return STATUS_SUCCESS?

sirknightj, line 154: Can we also add a non-success case too?

src/utils/src/Thread.c
```diff
@@ -159,14 +176,17 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
...
     pthread_attr_t attr;
-    pAttr = &attr;
-    result = pthread_attr_init(pAttr);
-    CHK_ERR(result == 0, STATUS_THREAD_ATTR_INIT_FAILED, "pthread_attr_init failed with %d", result);
-    result = pthread_attr_setstacksize(&attr, THREAD_STACK_SIZE_ON_CONSTRAINED_DEVICE);
-    CHK_ERR(result == 0, STATUS_THREAD_ATTR_SET_STACK_SIZE_FAILED, "pthread_attr_setstacksize failed with %d", result);
-#endif
+    if (stackSize != 0) {
+        if (stackSize < (SIZE_T) PTHREAD_MIN_STACK_SIZE) {
+            stackSize = (SIZE_T) PTHREAD_MIN_STACK_SIZE;
```
sirknightj, line 182: I would recommend logging or exit with some exception if given a value lower than the min.

tst/utils/Thread.cpp
jdelapla: Won't do

tst/utils/Thread.cpp
jdelapla: These functions are wrappers for operating system level functions. Writing tests that verify the range or validity of the OS function are unnecessary.

However I'll add one for the null check

jdelapla: Same as other comment.

src/utils/src/Thread.c
jdelapla: There seems to be a difference of opinion on how to tackle this error. Either let the pthread call fail and log it, or correct it to the allowed minimum.

There are valid points for both, I'll stick with the current implementation.

tst/utils/Thread.cpp
jdelapla: I'll add a rand()

README.md
jdelapla: done

CMakeLists.txt
disa6302: Yeah. 
The reason for the ask was to provide easy visibility for anyone looking into CMake file for a list of options. But, if we have documentation on how to set this up, I am fine not having this.

src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -1032,12 +1034,13 @@ extern PUBLIC_API atomicXor globalAtomicXor;
...
 // Thread functionality
 //
-#define THREAD_CREATE      globalCreateThread
-#define THREAD_JOIN        globalJoinThread
-#define THREAD_SLEEP       globalThreadSleep
-#define THREAD_SLEEP_UNTIL globalThreadSleepUntil
-#define THREAD_CANCEL      globalCancelThread
-#define THREAD_DETACH      globalDetachThread
+#define THREAD_CREATE             globalCreateThread
+#define THREAD_CREATE_WITH_PARAMS globalCreateThreadWithParams
```
hassanctech, line 1038: There are some differing view points on how we should handle errors.  On one hand you can take the approach (which you've taken) that you're going to pass the stack size requested by the client straight to the lower level APIs (whatever platform specific ones there are) and just return/log an error if they fail. A downside to this approach is it makes testing the behavior, in case an unsupported value is provided for a particular platform, impossible without knowing what values may be valid/invalid for that particular platform. An alternative approach could test an array of values and in the tests we if/def per platform to have the correct assert (success vs failure) for that platform.  

If we are going to move forward with the understanding that we're not going to do any platform specific validation rather just provide the parameter to the underlying threading library and log any errors, then we should at least make that very clear in the documentation that this is the behavior of the API.  Specifically that we will attempt to use the facilities provided by the threading library for your platform to set the stack size if there are errors returned we will surface them via logs and/or on some platforms it may cause a crash.  If you wish to avoid this, is up to you (the client) to make sure you're specifying stack sizes which are supported on your platform.

disa6302 (dismissed):

hassanctech (approved):
The Windows CI tests weren't running at all the WebRTC C SDK PR due to an unrelated issue (couldn't find ZLIB so the build didn't finish) so we weren't able to see if this works there.  I pushed a commit to that branch and ran the CI and it's all good on Windows except for some file caching tests which have been resolved in another PR, so I'm now approving this PR.

disa6302 (approved):

stefankiesz (approved):

PR #245, opened 2024-02-13
Reset current index in case it overflows out of range

jdelapla:
*Description of changes:*
In cases where the current index has passed the head value we needed a check to reset the index position. Without this check, you would need to iterate until the index had cycled around via integer overflow on a 64-bit integer.


disa6302 (approved):

hassanctech (approved):
lgtm

src/view/src/ContentView.c
```diff
@@ -153,6 +153,15 @@ STATUS contentViewGetNext(PContentView pContentView, PViewItem* ppItem)
     // Quick check if any items exist - early return
     CHK((pRollingView->head != pRollingView->tail) && (pRollingView->current != pRollingView->head), STATUS_CONTENT_VIEW_NO_MORE_ITEMS);
 
+    // current is greater than the header & head isn't currently in an overflow state.
```
stefankiesz, line 156: "header" should be "head"

PR #249, opened 2024-02-22
Fix readFile on Windows

niyatim23:
*Issue #, if available:*

*What was changed?*
- Fix readFile to work properly for multiline files on Windows

*Why was it changed?*
- The readFile was not working as expected from WebRTC C SDK in ASCII mode. To ensure that fread, fseek and ftell works on Windows as expected, compatible functions with expected return values are evaluated and used.

*How was it changed?*
- Use Windows-specific _fseeki64 and _ftelli64 as the traditional fseek and ftell are non-compliant on systems that do not provide the same guarantees as POSIX. On these systems, setting the file position indicator to the end of the file using fseek() is not guaranteed to work for a binary stream, and consequently, the amount of memory allocated may be incorrect, leading to a potential vulnerability ([Reference](https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek()+and+ftell()+to+compute+the+size+of+a+regular+file))
- fread would either return 1, i.e, the number of objects we've requested it to read or it would run into end-of-file / error. fread does not distinguish between end-of-file and error, so I used feof to determine it is indeed end-of-file and not an error. ([Reference](https://en.cppreference.com/w/c/io/fread))

- Wrote unit and functionality tests
- Tested manually on Windows
Screenshot 2024-02-22 at 8 58 01 AM
Screenshot 2024-02-22 at 8 57 47 AM
Screenshot 2024-02-22 at 8 57 25 AM

- Tested manually on MacOS
Screenshot 2024-02-22 at 1 43 25 PM

Screenshot 2024-02-22 at 8 54 11 AM

Screenshot 2024-02-22 at 8 53 52 AM

### DIFF
--- a/src/utils/src/FileIo.c
+++ b/src/utils/src/FileIo.c
@@ -23,8 +23,13 @@ STATUS readFile(PCHAR filePath, BOOL binMode, PBYTE pBuffer, PUINT64 pSize)
     CHK(fp != NULL, STATUS_OPEN_FILE_FAILED);
 
     // Get the size of the file
+#if defined _WIN32 || defined _WIN64
+    _fseeki64(fp, 0, SEEK_END);
+    fileLen = _ftelli64(fp);
+#else
     FSEEK(fp, 0, SEEK_END);
     fileLen = FTELL(fp);
+#endif
 
     if (pBuffer == NULL) {
         // requested the length - set and early return
@@ -75,8 +80,13 @@ STATUS readFileSegment(PCHAR filePath, BOOL binMode, PBYTE pBuffer, UINT64 offse
     CHK(fp != NULL, STATUS_OPEN_FILE_FAILED);
 
     // Get the size of the file
+#if defined _WIN32 || defined _WIN64
+    _fseeki64(fp, 0, SEEK_END);
+    fileLen = _ftelli64(fp);
+#else
     FSEEK(fp, 0, SEEK_END);
     fileLen = FTELL(fp);
+#endif
 
     // Check if we are trying to read past the end of the file
     CHK(offset + readSize <= fileLen, STATUS_READ_FILE_FAILED);
--- a/src/utils/src/FileIo.c
+++ b/src/utils/src/FileIo.c
@@ -80,7 +80,7 @@ STATUS readFileSegment(PCHAR filePath, BOOL binMode, PBYTE pBuffer, UINT64 offse
     CHK(fp != NULL, STATUS_OPEN_FILE_FAILED);
 
     // Get the size of the file
-#if defined _WIN32 || defined _WIN64
+#ifdef _WIN32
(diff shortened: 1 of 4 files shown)


hassanctech:
Nice!  Let's add some unit test coverage here for this bug fix, one that fails before the change but succeeds now.

disa6302:
Awesome!
In addition to unit test, are there unit tests in WebRTC we should remove from the gtest filter to test file related operations (I think the file caching tests in there are disabled at the moment for windows) ? If so, can we open a PR in webrtc referencing this commit, enabling the file cache specific tests to make sure it is all good?

niyatim23:
> Awesome! In addition to unit test, are there unit tests in WebRTC we should remove from the gtest filter to test file related operations (I think the file caching tests in there are disabled at the moment for windows) ? If so, can we open a PR in webrtc referencing this commit, enabling the file cache specific tests to make sure it is all good?

Enabled Signaling tests on Windows in: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1933

src/utils/src/FileIo.c
```diff
@@ -23,8 +23,18 @@ STATUS readFile(PCHAR filePath, BOOL binMode, PBYTE pBuffer, PUINT64 pSize)
     CHK(fp != NULL, STATUS_OPEN_FILE_FAILED);
 
     // Get the size of the file
+
+    // Use Windows-specific _fseeki64 and _ftelli64 as the traditional fseek and ftell are non-compliant on systems that do not provide
+    // the same guarantees as POSIX. On these systems, setting the file position indicator to the
+    // end of the file using fseek() is not guaranteed to work for a binary stream, and consequently,
+    // the amount of memory allocated may be incorrect, leading to a potential vulnerability.
+#if defined _WIN32 || defined _WIN64
+    _fseeki64(fp, 0, SEEK_END);
```
hassanctech, line 32: Is it the case for Windows that we should *always* use `_fseeki64` instead of `fseek` regardless of arguments?  If so I think we should adjust the #define here:

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/57637ea593f4b43c509413a44d993ed08d7f2616/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L900

Witha n if/def for Windows to have 

```
#ifndef FSEEK
#if defined _WIN32 || defined _WIN64
#define FSEEK _fseeki64
#else
#define FSEEK fseek
#endif
#endif

hassanctech: Same for `FTELL`

niyatim23: Done!

hassanctech (approved):
LGTM

disa6302 (approved):

PR #251, opened 2024-03-01
Change fragment metadata count from 25 to 10

niyatim23:
*Issue #, if available:*

*What was changed?*
- Modified `MAX_FRAGMENT_METADATA_COUNT` to 10 
- Modified actions/checkout@v2 to v4
- Moved from macos-11 to macos-12 as the former has been deprecated

*Why was it changed?*
- Changed because that is the limit from the backend
- v2 is an older version and v4 is the latest

*How was it changed?*
- Updated the `MAX_FRAGMENT_METADATA_COUNT`
- Updated the ci.yml

- The CI tests cover this change

### DIFF
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -272,7 +272,7 @@ extern "C" {
 /**
  * Maximal size of the metadata queue for a fragment
  */
-#define MAX_FRAGMENT_METADATA_COUNT 25
+#define MAX_FRAGMENT_METADATA_COUNT 10
 
 /**
  * Maximum amount of the "tags" element metadata for a fragment
--- a/src/client/include/com/amazonaws/kinesis/video/client/Include.h
+++ b/src/client/include/com/amazonaws/kinesis/video/client/Include.h
@@ -277,7 +277,7 @@ extern "C" {
 /**
  * Maximum amount of the "tags" element metadata for a fragment
  */
-#define MAX_FRAGMENT_METADATA_TAGS 10
+#define MAX_FRAGMENT_METADATA_TAGS MAX_FRAGMENT_METADATA_COUNT
 
 /**
  * Max name/value pairs for custom event metadata
--- a/tst/client/StreamPutGetTest.cpp
+++ b/tst/client/StreamPutGetTest.cpp
@@ -1427,8 +1427,8 @@ TEST_F(StreamPutGetTest, putFrame_PutGetNotifyAndTagsStoreData)
         // Set the frame bits
         MEMSET(frame.frameData, (BYTE) i, SIZEOF(tempBuffer));
 
-        // Key frame every 10th
-        frame.flags = i % 10 == 0 ? FRAME_FLAG_KEY_FRAME : FRAME_FLAG_NONE;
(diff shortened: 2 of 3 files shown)


tst/client/StreamPutGetTest.cpp
```diff
@@ -1427,8 +1427,8 @@ TEST_F(StreamPutGetTest, putFrame_PutGetNotifyAndTagsStoreData)
         // Set the frame bits
         MEMSET(frame.frameData, (BYTE) i, SIZEOF(tempBuffer));
 
-        // Key frame every 10th
-        frame.flags = i % 10 == 0 ? FRAME_FLAG_KEY_FRAME : FRAME_FLAG_NONE;
```
stefankiesz, line 1431: Would like to be sure there isn't a reason we had it at 10 frames per fragment, but hard to know without OC's input. If there is a reason and we want to keep the 10 frames/fragment, can skip the first 2 frames from getting metadata added.

stefankiesz (approved):

tst/client/StreamPutGetTest.cpp
niyatim23: From the test it doesn't seem like there was a reason for that. Previously there were 10 key frames and each fragment was allowed 25 metadata tags max. Now, it is 8 frames per fragment and they're allowed 10 metadata tags. I had to keep it 8 because putEventMetadata adds 2 tags per call and if the keyframe is every 10th frame, we exceed the max tags which is 10 now

sirknightj (approved):

sirknightj:
https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/limits.html#limits-streaming-metadata

src/client/include/com/amazonaws/kinesis/video/client/Include.h
```diff
@@ -272,7 +272,7 @@ extern "C" {
 /**
  * Maximal size of the metadata queue for a fragment
  */
-#define MAX_FRAGMENT_METADATA_COUNT 25
+#define MAX_FRAGMENT_METADATA_COUNT 10
 
 /**
  * Maximum amount of the "tags" element metadata for a fragment
```
hassanctech, line 278: Can we delete `MAX_FRAGMENT_METADATA_TAGS` or use that instead since it appears it's not being used otherwise ?

niyatim23: I did not want to do that since the values are publicly available, some customer might be using them

disa6302: I agree with not removing because it is a public define. But I am also wondering if it would be worth just setting it this way:
`#define MAX_FRAGMENT_METADATA_COUNT MAX_FRAGMENT_METADATA_TAGS ` or, set it the other way round:

`#define MAX_FRAGMENT_METADATA_TAGS MAX_FRAGMENT_METADATA_COUNT`

To make sure in the event customers are using `MAX_FRAGMENT_METADATA_TAGS`, they are reading the right value.

niyatim23: Done!

hassanctech (approved):
lgtm

disa6302 (approved):

PR #258, opened 2024-10-23
Enhance stream state recovery after interruptions

PoshSylvester:
Description of changes: This PR introduces improvements in the handling of stream state recovery in cases of interruptions, such as application crashes or connection drops, to prevent loss of important session data. Specifically, the changes include:

Stream Recovery State Structure: Added a new StreamRecoveryState structure to track essential stream properties like the current view item index, timestamps (PTS/DTS), transferred bytes, and flags indicating whether metadata is being sent (e.g., eosMetadataSendFlag). This allows for more precise recovery of the stream state after interruptions.

Improved State Persistence: The saveStreamRecoveryState and loadStreamRecoveryState functions have been updated to save the stream state at key frames (start of each fragment). This ensures that when an interruption occurs, the stream can resume from the last known good state.

Connection Failure Handling: Added better detection of interruptions using connectionDroppedResult and connectionState flags. If an error or crash is detected, the stream state is loaded from persistent storage and the streaming session resumes without losing critical session information.

Testing:

Verified the changes by simulating application crashes and connection drops. Ensured that streams recover and continue streaming from the last saved key frame without any data loss.

### DIFF
--- a/src/client/src/Stream.c
+++ b/src/client/src/Stream.c
@@ -1157,6 +1157,79 @@ CleanUp:
     return retStatus;
 }
 
+
+/**
+ * Saves the StreamRecoveryState to a binary file.
+ * 
+ * @param state The state to save.
+ * @return STATUS_SUCCESS if successful, an error status otherwise.
+ */
+STATUS saveStreamRecoveryState(StreamRecoveryState state)
+{
+    FILE *file = fopen(RECOVERY_STATE_FILE_PATH, "wb");
+    if (file == NULL) {
+        return STATUS_PERSISTENCE_ERROR;
+    }
+
+    size_t written = fwrite(&state, sizeof(StreamRecoveryState), 1, file);
+    if (written != 1) {
+        fclose(file);
+        return STATUS_PERSISTENCE_ERROR;
+    }
+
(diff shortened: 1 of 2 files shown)


MushMal (commented):
I added a few comments regarding the codebase but the biggest concern is that this is not going to do anything for crash recovery as preserving just a small portion of the state will not let anything to recover.

I think there is a good reason for some sort of recovery strategy but it needs to be holistic and address various scenarios for crash recovery which is not trivial and likely needs to be driven by the application itself. 

Perhaps, sharing the actual scenario might prompt other contributors to come up with an existing higher-level solution or similar. Or, if this is a unique case, a more holistic approach  in the design might be required.

PoshSylvester:
Thank you for your feedback, @MushMal. Based on your suggestions, we’re pivoting to handle state persistence and recovery directly within the application rather than embedding it into the SDK. This new approach would make the application responsible for orchestrating the recovery process, allowing for more flexibility and adaptability to diverse use cases. To support this, we’ll  need to introduce two  SDK functions to get and set the stream state as needed. This approach should also address concerns about code portability, optimization, and performance across different stream types.

PR #262, opened 2024-11-01
Apply linting

sirknightj:
*Issue #, if available:*
N/A

*What was changed?*
- Code style was changed to fit the repository's current linter rules.
- Previously, linter was only applied to the `src/` directory. It is now applied to the `tst/` directory too.
- Note: I think some of the repository's linter configurations should be adjusted. We can revisit that in a future pull request.

*Why was it changed?*
- To have a consistent code formatting.

*How was it changed?*
- Clang-format patch was applied.

- The clang-format check is passing for this: https://github.com/awslabs/amazon-kinesis-video-streams-pic/actions/runs/11624955634


unicornss (approved):
lint changesw

PR #264, opened 2024-11-05
Custom Thread StackSize API

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- See #243 for the original PR with details.
- Add new CMake flag (`KVS_DEFAULT_STACK_SIZE`) that allows users to configure the default stack size of threads created with `THREAD_CREATE`.
- Add new CMake flag (`THREAD_CREATE_WITH_PARAMS`) that allows users to create threads with custom stack sizes.
- For non-windows, there was a CMake flag available: `-DCONSTRAINED_DEVICE` that sets the default stack size to `0.5 MiB`. For backwards compatibility, this flag was not changed.
- Make `KVS_DEFAULT_STACK_SIZE` and  `CONSTRAINED_DEVICE` incompatible with each other. The build will fatally exit if both of these are set.
- Added more unit tests for the new API, including a unit test that checks that validates the stack size was actually changed.
- Add a script to run various CMake flag combinations to check that that it correctly fails the build if those 2 exclusive CMake flags are set. Generate a markdown test report for that as well.

*Why was it changed?*
* See #243 for additional details. 

*How was it changed?*
* See #243 for additional details. 

* See #243 for additional details. 
* For the new changes, the newly added tests are passing and the new test report can be found at the bottom of the page: https://github.com/awslabs/amazon-kinesis-video-streams-pic/actions/runs/11676076021
image


tst/utils/Thread.cpp
```diff
@@ -246,7 +246,7 @@ TEST_F(ThreadFunctionalityTest, VerifyStackSize)
     EXPECT_NE(0, threadId2);
     EXPECT_EQ(STATUS_SUCCESS, THREAD_JOIN(threadId2, NULL));
     EXPECT_EQ(0, threadInfo2.failure);
-    EXPECT_EQ(threadStack, threadInfo2.stackSize);
+    EXPECT_EQ(threadStack2, threadInfo2.stackSize);
```
unicornss, line 249: Can   ` threadStack2` be named as  `oneMiBStackSize` and `threadInfo2` --> `oneMiBThreadInfo` ?

unicornss (approved):

tst/utils/Thread.cpp
sirknightj: I can change the name of the variable, yes.

src/utils/src/Thread.c
```diff
@@ -198,6 +213,23 @@ PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID
...
 }
 
+PUBLIC_API STATUS defaultCreateThread(PTID pThreadId, startRoutine start, PVOID args)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+#if defined(KVS_DEFAULT_STACK_SIZE_BYTES)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, (SIZE_T) KVS_DEFAULT_STACK_SIZE_BYTES, args));
+#elif defined(CONSTRAINED_DEVICE)
+    CHK_STATUS(defaultCreateThreadWithParams(pThreadId, start, (SIZE_T) THREAD_STACK_SIZE_ON_CONSTRAINED_DEVICE, args));
```
stefankiesz, line 223: Although there is a check for the case of KVS_DEFAULT_STACK_SIZE_BYTES and CONSTRAINED_DEVICE both being defined through the CMakeLists, it is still possible to set both of these elsewhere such as #defines, in which case there is potential for hidden un-intended behavior. Can we consider adding a similar check within any of these if-defs?

tst/utils/Thread.cpp
```diff
@@ -113,3 +123,130 @@ TEST_F(ThreadFunctionalityTest, ThreadCreateAndCancel)
...
+    INT32 failure;
+} TestThreadInfo;
+
+// Function to retrieve and print the stack size from within the thread
+PVOID fetchStackSizeThreadRoutine(PVOID arg)
+{
+    pthread_attr_t attr;
+    SIZE_T stackSize;
+    INT32 result = 0;
+    TestThreadInfo* pThreadInfo = (TestThreadInfo*) arg;
```
stefankiesz, line 204: Do we want to null check this?

stefankiesz (approved):

tst/utils/Thread.cpp
sirknightj: Added just in case

MushMal (commented):
This is a good change. My only concern is the name of the new API and the choice of the parameter itself. What if in the future you might want to add another parameter? I would suggest adding a struct that can be versioned and new params could be added later for back compat. The struct initially can have only a couple of fields - like the version, stack size, thread name. But later can include more flags such as start suspended, priority, etc..

unicornss:
Something along these lines ... 
```
#define THREAD_PARAMS_VERSION_1 1
typedef struct ThreadParams {
    int version;
    SIZE_T stackSize;
    :
} ThreadParams;

```

PR #273, opened 2025-01-28
Update UINT32 and SIZE_T Definitions

stefankiesz:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/2065

*What was changed?*
- The defining of `UINT32` and `SIZE_T`.
- Added a CI test to check for UINT32 actually being of size 32 bits.

*Why was it changed?*
- On some platforms, `UINT32` is improperly being set to a 64-bit integer.
  - This was causing the definition of `SIZE_T` (which, on some platforms, we set to `UINT32`) to not correspond to `size_t`, which caused parameter type mismatch compilation errors.
- If `size_t` is available on the system, then we should define `SIZE_T` as such.

*How was it changed?*
- Check whether `UINT_MAX` is equal to 32-bit max, if it is, set UINT32 to `int` instead of `long`.
- Added check for `size_t` definition before defining `SIZE_T`.

- Allowing for KVS PIC CI to pass.
- Allowing for the CI to pass on the dependent SDKs:
 - Producer C
 - Producer C++
 - WebRTC-C
- Tested the new unit test by forcing a failure case. 


src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h
```diff
@@ -145,8 +145,8 @@ typedef unsigned char UINT8;
 typedef char INT8;
 typedef unsigned short UINT16;
 typedef short INT16;
-typedef unsigned long UINT32;
-typedef long INT32;
+typedef unsigned int UINT32;
+typedef int INT32;
```
hassanctech, line 153: The C standard only guarantees that an `int` be at least 16 bits so this can fail on some platforms.

stefankiesz: Thanks for calling this out, I added a conditional to check for the size of `UINT_MAX` which comes from `limits.h` which, from my findings, should be available on all platforms.

vikramdattu: Ideally, if needed needed, INT32 should define to `int32_t` from `inttypes.h` guaranteeing it to be always 32 bits and likewise for other macros which have exact sizes.

PR #277, opened 2025-02-12
Adding more chk log err's

sirknightj:
*Issue #, if available:*
N/A

*What was changed?*
- Adding more CHK_LOG_ERR.

*Why was it changed?*
- C has errors returned by value. This should make debugging easier to pinpoint error locations, especially in environments where gdb isn't available to generate a stacktrace.

*How was it changed?*
- Added chk_log_err in the cleanup label of functions related to the Streams state machine, semaphores, and mkvgenerator.

- Ci should pass for these changes.

### DIFF
--- a/src/utils/src/Semaphore.c
+++ b/src/utils/src/Semaphore.c
@@ -16,6 +16,7 @@ STATUS semaphoreCreate(UINT32 maxPermits, PSEMAPHORE_HANDLE pHandle)
     *pHandle = TO_SEMAPHORE_HANDLE(pSemaphore);
 
 CleanUp:
+    CHK_LOG_ERR(retStatus);
 
     if (STATUS_FAILED(retStatus)) {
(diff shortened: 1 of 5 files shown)


src/utils/src/Semaphore.c
```diff
@@ -385,6 +389,8 @@ STATUS semaphoreGetCountInternal(PSemaphore pSemaphore, PINT32 pCount)
     *pCount = (INT32) ATOMIC_LOAD(&pSemaphore->permitCount);
 
 CleanUp:
+    CHK_LOG_ERR(retStatus);
```
hassanctech, line 392: This is going to spam a LOT of error logs for many instances where returning a non-success is completely benign / not an error at all.  IMO it makes much more sense for the client application or higher level data structure that is using to determine whether what's returned from here is an error and let it decide if it wants to bail out early and log the error.

PR #283, opened 2025-07-24
Use the kvs retry strategy config that was passed in

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Fixed retry strategy configuration persistence during client initialization by adding missing field copies in `fixupClientInfo()`. The retry strategy configuration (`kvsRetryStrategy`) and callbacks (`kvsRetryStrategyCallbacks`) are now properly preserved when client info undergoes version compatibility processing.

*Why was it changed?*
- The retry strategy configuration system was effectively non-functional due to a silent bug in the client initialization path. When users provided custom retry configurations (exponential backoff parameters, jitter settings, etc.), these settings were being silently discarded during the `fixupClientInfo()` validation process, causing all clients to fall back to hardcoded defaults.

> [!Note]
> The retry strategy configuration is only for configuring **retry times** (how long to wait between attempts), it does not impact the retry behavior or number of attempts to retry a certain error.

- Why did the previous test miss it?
  - The previous test overrides the default values in the constructor clientInfo.kvsRetryStrategyCallbacks.createRetryStrategyFn, so the "real" constructor wasn't actually verified to preserve the values.
- Also add some more comments to explain that the units are in milliseconds (anti-pattern since the rest of the code uses hundreds of nanos units...) 

*How was it changed?*
- Added missing MEMCPY in `src/client/src/InputValidator.c:fixupClientInfo()` when copying over the original value from the user.
- Enhanced error handling, debugging logs in `src/utils/src/ExponentialBackoffRetryStrategy.c` since there are max and min values for the fields. (Improved configuration validation feedback with debug messages for each validation step)

- Created `tst/client/InputValidatorTest.cpp` with targeted test cases:
   - Verifies retry strategy config and callbacks are properly copied
   - Tests the exact code path that was broken
- Created `tst/client/RetryStrategyConfigurationTest.cpp` that:
  - Calls the createKinesisVideoClient constructor API with the deviceInfo and clientInfo, and validates the retry config is preserved.

### DIFF
--- a/tst/client/RetryStrategyConfigurationTest.cpp
+++ b/tst/client/RetryStrategyConfigurationTest.cpp
@@ -1,0 +1,195 @@
+#include "ClientTestFixture.h"
+
+class RetryStrategyConfigurationTest : public ClientTestBase {
+  public:
+    void setupCustomRetryConfiguration(PClientInfo pClientInfo, PExponentialBackoffRetryStrategyConfig pCustomConfig) {
+        // Set up custom retry configuration values that are different from defaults
+        // Values are in MILLISECONDS (will be converted internally to hundreds of nanos)
+        pCustomConfig->maxRetryCount = 10; // Different from default KVS_INFINITE_EXPONENTIAL_RETRIES
+        pCustomConfig->maxRetryWaitTime = 15000; // 15 seconds (min is 10000ms)
+        pCustomConfig->retryFactorTime = 100; // 100ms (min is 50ms)
+        pCustomConfig->minTimeToResetRetryState = 120000; // 120 seconds (min is 90000ms)
+        pCustomConfig->jitterType = FIXED_JITTER; // Different from default FULL_JITTER
+        pCustomConfig->jitterFactor = 100; // Custom jitter factor (min is 50)
+
+        // Set up retry strategy with custom config
+        pClientInfo->kvsRetryStrategy.pRetryStrategyConfig = (PRetryStrategyConfig) pCustomConfig;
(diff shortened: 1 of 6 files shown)


Ahhorse1 (approved):

tst/client/RetryStrategyConfigurationTest.cpp
```diff
@@ -0,0 +1,317 @@
...
+    deviceInfo.clientInfo.kvsRetryStrategyCallbacks.executeRetryStrategyFn = getExponentialBackoffRetryStrategyWaitTime;
+    deviceInfo.clientInfo.kvsRetryStrategyCallbacks.getCurrentRetryAttemptNumberFn = getExponentialBackoffRetryCount;
+
+    // Create client - this should FAIL because our validation will detect wrong config values
+    STATUS result = createKinesisVideoClient(&deviceInfo, &mClientCallbacks, &clientHandle);
+    EXPECT_EQ(STATUS_INVALID_ARG, result); // Should fail with our validation error
+
+    // If it somehow succeeded, clean up
+    if (result == STATUS_SUCCESS) {
+        freeKinesisVideoClient(&clientHandle);
```
stefankiesz, line 315: Want to confirm we shouldn't cleanup if failed?

stefankiesz (approved):

tst/client/RetryStrategyConfigurationTest.cpp
sirknightj: Correct. If the `createKinesisVideoClient` failed, then the function itself takes care of cleaning it up - the returned clientHandle is not valid. Hence, the caller should not call free on an invalid handle.

PR #290, opened 2026-01-07
Add environment variable check utility function

stefankiesz:
*Issue #, if available:*
N/A

*What was changed?*
- Added a new utility function to check whether an environment variable is enabled. Checks like this are made throughout consuming SDKs, so adding a utility function to centralize it and keep logic consistent.
- Changed verbosity of "Opening file" logs to VERBOSE to reduce log clutter in DEBUG mode.

The function returns TRUE if the passed in env var is set to 1, "ON", or "OFF" (case insensitive), else returns FALSE.

*Why was it changed?*
To provide a helper function for all down-stream libraries to be able to use.

*How was it changed?*
Created a new _Environment.c_ source file and added the definition there.

- Added new unit tests to cover various cases.
- CI to pass.

### DIFF
--- a/src/utils/src/Environment.c
+++ b/src/utils/src/Environment.c
@@ -15,7 +15,7 @@ BOOL isEnvVarEnabled(PCHAR envVarName)
     retBool = envVarVal != NULL && (STRCMPI(envVarVal, "1") == 0 || STRCMPI(envVarVal, "true") == 0 || STRCMPI(envVarVal, "on") == 0);
 
 CleanUp:
-    UNUSED_PARAM(retStatus);
+    CHK_LOG_ERR(retStatus);
     LEAVES();
     return retBool;
 }
--- a/src/utils/src/Environment.c
+++ b/src/utils/src/Environment.c
@@ -1,0 +1,21 @@
+#include "Include_i.h"
+
+BOOL isEnvVarEnabled(PCHAR envVarName)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL retBool = FALSE;
+
+    // Null or empty envVarName to GETENV is undefined behavior.
+    CHK_ERR(envVarName != NULL && envVarName[0] != '\0', STATUS_NULL_ARG, "Environment variable name is NULL or empty.");
+
+    PCHAR envVarVal = GETENV(envVarName);
+
+    // Case-insensitive comparisons.
+    retBool = envVarVal != NULL && (STRCMPI(envVarVal, "1") == 0 || STRCMPI(envVarVal, "true") == 0 || STRCMPI(envVarVal, "on") == 0);
+
+CleanUp:
+    UNUSED_PARAM(retStatus);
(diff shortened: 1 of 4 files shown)


Ahhorse1 (approved):
Seems to be a lot of small formatting changes as well, should we separate that from this PR?

src/utils/src/Environment.c
```diff
@@ -0,0 +1,9 @@
+#include "Include_i.h"
+
+BOOL isEnvVarEnabled(PCHAR envVarName)
+{
+    PCHAR envVarVal = GETENV(envVarName);
```
sirknightj, line 12: Can we add a null-check for the `envVarName` for extra safety?

And the corresponding test

sirknightj (approved):

.github/workflows/lint.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     - name: Check clang
       run: |
-        find . -name "*.c" -o -name "*.h" | xargs clang-format -style=file --dry-run -Werror
+        find . -name "*.c" -o -name "*.h" | xargs clang-format-14 -style=file --dry-run -Werror
```
sirknightj: This should already be handled by line 13?

sirknightj (approved):

src/utils/src/Environment.c
```diff
@@ -0,0 +1,21 @@
...
+    // Null or empty envVarName to GETENV is undefined behavior.
+    CHK_ERR(envVarName != NULL && envVarName[0] != '\0', STATUS_NULL_ARG, "Environment variable name is NULL or empty.");
+
+    PCHAR envVarVal = GETENV(envVarName);
+
+    // Case-insensitive comparisons.
+    retBool = envVarVal != NULL && (STRCMPI(envVarVal, "1") == 0 || STRCMPI(envVarVal, "true") == 0 || STRCMPI(envVarVal, "on") == 0);
+
+CleanUp:
+    UNUSED_PARAM(retStatus);
```
sirknightj: Should we CHK_LOG_ERR?

sirknightj (approved):

src/utils/src/FileIo.c
```diff
@@ -19,7 +19,7 @@ STATUS readFile(PCHAR filePath, BOOL binMode, PBYTE pBuffer, PUINT64 pSize)
 
     CHK(filePath != NULL && pSize != NULL, STATUS_NULL_ARG);
 
-    DLOGD("Opening file: %s", filePath);
+    DLOGV("Opening file: %s", filePath);
```
Ahhorse1, line 22: Is there a reason we want these to be verbose instead of debug?

Ahhorse1 (approved):

src/utils/src/FileIo.c
stefankiesz: Yes, it prints very many logs when opening video frame files.

sirknightj (approved):

PR #294, opened 2026-03-26
Add isEnvEnabled function with default when env is unset

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*

Added `isEnvVarEnabledWithDefault(PCHAR, BOOL)` utility function alongside the existing `isEnvVarEnabled`.

*Why was it changed?*

`isEnvVarEnabled` supports default-OFF semantics. Default-ON environment variables were not supported. There was no clean way to check if a user explicitly disabled a feature that is enabled by default.

<details><summary><strong>Detailed explanation</strong></summary>

For example, given a feature that is ON by default, and the user decided to set it to on anyway:
```bash
export USE_FEATURE=ON
```

```c
// Incorrect: disabled even though the user explicitly enabled it
if (!isEnvVarEnabled("USE_FEATURE")) { ... }
```

Before `!isEnvVarEnabled()`:

| USE_FEATURE | Result | Correct? |
|---|---|---|
| ON | FALSE | ❌ Feature disabled despite user enabling it |
| OFF | TRUE | ❌ Feature enabled despite user disabling it |
| unset | TRUE | ✅ |

After `isEnvVarEnabledWithDefault(env, TRUE)`:

| USE_FEATURE | Result | Correct? |
|---|---|---|
| ON | TRUE | ✅ |
| OFF | FALSE | ✅ |
| unset | TRUE | ✅ Default preserved |

</details>

*How was it changed?*

- Introduced `isEnvVarEnabledWithDefault(PCHAR envVarName, BOOL defaultVal)` to return the `defaultVal` when the env var is unset.
  - Note: In C, there is no method overloading, so the new function needed a different name.
- Reimplemented `isEnvVarEnabled` as a thin wrapper: `isEnvVarSet(envVarName, FALSE)` for backward compatibility.

- Added new unit tests to check the new functionality
- Existing tests for `isEnvVarEnabled` continue to pass (backwards compatibility)

```bash
 ./tst/kvspic_test --gtest_filter="*EnvironmentFunctionalityTest*"
...
[       OK ] EnvironmentFunctionalityTest.isEnvVarEnabledWithDefaultNullAndEmpty (0 ms)
[----------] 7 tests from EnvironmentFunctionalityTest (2 ms total)

[----------] Global test environment tear-down
[==========] 7 tests from 1 test suite ran. (3 ms total)
[  PASSED  ] 7 tests.
```

### DIFF
--- a/tst/utils/Environment.cpp
+++ b/tst/utils/Environment.cpp
@@ -18,7 +18,7 @@ TEST_F(EnvironmentFunctionalityTest, CheckTrueCases)
     for (std::string val : trueValues) {
 // Unset env var first.
 #ifdef _WIN32
-        SNPRINTF(envBuf, sizeof(envBuf), "%s=", envVar.c_str());
+        SNPRINTF(envBuf, SIZEOF(envBuf), "%s=", envVar.c_str());
         _putenv(envBuf);
 #else
         unsetenv(envVar.c_str());
@@ -29,7 +29,7 @@ TEST_F(EnvironmentFunctionalityTest, CheckTrueCases)
 
 // Set to the TRUE value.
 #ifdef _WIN32
-        SNPRINTF(envBuf, sizeof(envBuf), "%s=%s", envVar.c_str(), val.c_str());
+        SNPRINTF(envBuf, SIZEOF(envBuf), "%s=%s", envVar.c_str(), val.c_str());
         _putenv(envBuf);
 #else
         setenv(envVar.c_str(), val.c_str(), 1);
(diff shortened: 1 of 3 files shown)


tst/utils/Environment.cpp
```diff
@@ -18,7 +18,7 @@ TEST_F(EnvironmentFunctionalityTest, CheckTrueCases)
     for (std::string val : trueValues) {
 // Unset env var first.
 #ifdef _WIN32
-        SNPRINTF(envBuf, sizeof(envBuf), "%s=", envVar.c_str());
+        SNPRINTF(envBuf, SIZEOF(envBuf), "%s=", envVar.c_str());
```
Ahhorse1, line 21: Is there a reason for replacing it with the macro instead for all of these?

Ahhorse1 (approved):

tst/utils/Environment.cpp
sirknightj: The macro gives the dev the option to override the behavior.

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/a9ec1d202750d41188922590a779c9b9d213ba14/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L394-L396

YPandas (approved):

PR #296, opened 2026-07-09
notify streamErrorReportFn calbacks when call result is permission error

vabhasin:
*Issue #, if available:*

- N/A

*What was changed?*

- In `streamTerminatedEvent()`, added a call to `streamErrorReportFn` when the call result is `SERVICE_CALL_FORBIDDEN` (403) or `SERVICE_CALL_NOT_AUTHORIZED` (401).

*Why was it changed?*

- When PutMedia returns HTTP 403 (e.g., IAM role lacks `kinesisvideo:PutMedia` permission), the error was never reported to the application. `streamTerminatedEvent` only drove the state machine retry cycle — it never invoked `streamErrorReportFn`. This meant:
  - The application's error handler (e.g., kvssink's `streamErrorReportHandler`) was never called
  - The application had no way to detect or react to a permanent permission failure
  - The state machine retried infinitely without the caller knowing why
- We still keep the PIC behavior of retrying indefinitely on this error unchanged, while providing a way to registered handlers to do something about the permission error (reset stream, stop pipeline etc).
*How was it changed?*

- Added a check at the top of `streamTerminatedEvent()` (before the state lock is acquired) that detects 403/401 call results and invokes `streamErrorReportFn` with `STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR`.
- The state machine continues to retry (with backoff). This change only adds notification, it does not alter the retry behavior. The application can choose to stop the stream or log the error based on its own policy.
- The notification is placed before the lock intentionally as callbacks must not block while holding the stream mutex.

- Unit tests pass (existing `StreamApiFunctionalityTest` suite)
- End to end manual test: removed PutMedia permission from IAM role, verified `streamErrorReportFn` is called with `STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR` on first 403
- Verified the callback chain: ContinuousRetryHandler passes through (returns SUCCESS), application handler receives the error
- Verified no behavior change for other error types (500, timeouts still retry normally)

### DIFF
--- a/src/client/src/StreamEvent.c
+++ b/src/client/src/StreamEvent.c
@@ -655,6 +655,17 @@ STATUS streamTerminatedEvent(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HAN
     CHK(pKinesisVideoStream != NULL && pKinesisVideoStream->pKinesisVideoClient != NULL, STATUS_NULL_ARG);
     pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
 
+    // Report permission errors to the registered stream error handlers
+    // PIC state machine will still retry but application must be informed to handle the error appropriately.
+    if (callResult == SERVICE_CALL_FORBIDDEN || callResult == SERVICE_CALL_NOT_AUTHORIZED) {
+        DLOGE("[%s] Stream terminated with authorization error (result=%u). Notifying application.", pKinesisVideoStream->streamInfo.name, callResult);
+        if (pKinesisVideoClient->clientCallbacks.streamErrorReportFn != NULL) {
+            pKinesisVideoClient->clientCallbacks.streamErrorReportFn(pKinesisVideoClient->clientCallbacks.customData,
+                                                                     TO_STREAM_HANDLE(pKinesisVideoStream), uploadHandle, 0,
+                                                                     STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR);
+        }
+    }
+
     // Lock the state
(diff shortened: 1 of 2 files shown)


src/client/src/StreamEvent.c
```diff
@@ -655,6 +655,18 @@ STATUS streamTerminatedEvent(PKinesisVideoStream pKinesisVideoStream, UPLOAD_HAN
     CHK(pKinesisVideoStream != NULL && pKinesisVideoStream->pKinesisVideoClient != NULL, STATUS_NULL_ARG);
     pKinesisVideoClient = pKinesisVideoStream->pKinesisVideoClient;
 
+    // Report permission errors to the registered stream error handlers
+    // PIC state machine will still retry but application must be informed to handle the error appropriately.
+    if (callResult == SERVICE_CALL_FORBIDDEN || callResult == SERVICE_CALL_NOT_AUTHORIZED) {
```
sirknightj, line 660: Should there be other codes like 400?

e.g. `callResult != SERVICE_CALL_RESULT_OK`?

vabhasin: Yea we could do that, my original thought was that notifying on all non 200 would be a bit noisy, for eg. a transient 500 that recovers in 2 seconds, does not warrant an application callback and we already have PIC retrying for such errors.  
Another way to think would be that we know that the 403/401 case is non-recoverable and we directly tackle this one case.

Having said that, Im still okay to do that since the listener application can itself filter on these errors and act accordingly.

vabhasin: Also, I feel a better place to add this will be inside the `PutStreamCurlHandler`
https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/blob/master/src/source/CurlApiCallbacks.c#L2276-L2280

